PR #9, opened 2020-05-25
WIP: Create shex

matentzn:
With thanks to @lltommy 

Will work on this together over the next months. Thanks


cmungall:
Shall we move the schema folder from here?

https://github.com/cmungall/sssom-py/blob/master/schema/sssom.shex

matentzn:
I think this makes sense!

matentzn:
This has now been incorporated in the biolink model
https://github.com/mapping-commons/sssom-py/blob/master/schema/sssom.yaml

PR #27, opened 2020-08-10
Create sssom-example-catalog-0.0.1.yml (for discussion)

matentzn:
We are thinking for providing a little schema for combining sets of mapping sets to a catalog. This is just a first example for discussion. Note that this also introduces a `title` and a `data` property for the mapping sets.


matentzn:
This is now developed elsewhere:
https://github.com/mapping-commons/mapping-commons-template

PR #55, opened 2021-03-10
Clarify comment block in embedded mode

### LINKED ISSUE #54: Constraints on TSV comments
Is it the case that no comment lines can appear once the TSV starts? Meaning, the comments must be in a contiguous block at the start of the file? This would make sense to me. 

Also, is it the case that any comments must constitute valid YAML syntax after stripping the leading `#`? That's the impression I get from the spec, but I think it could be stated more directly.

matentzn:
fixes #54 

However, reviewing this made me realise that this spec does not consider the case when the hash symbol appears in a value. For example:

```
mapping_source: https://github.com/INCATools/ontology-development-kit/blob/master/docs/ReleaseArtefacts.md#release-artefact-2-full-required
```

or 

```
HP:001 skos:exactMatch MP:001  https://github.com/INCATools/ontology-development-kit/blob/master/docs/ReleaseArtefacts.md#release-artefact-2-full-required
```

Ideas @cmungall @balhoff ?

### DIFF
--- a/SSSOM.md
+++ b/SSSOM.md
@@ -328,6 +328,45 @@ HP:0000411	skos:exactMatch	MP:0000021	Lexical	Protruding ear	prominent ears
 
 In the embedded mode, we allow the integration of mapping set level metadata as **_commented YAML_**. Apart from being commented, the YAML follows the exact same spec as the *YAML specified by the external mode*. Heavily used tools in bioinformatics such as pandas allow to [specify comment characters](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.read_csv.html) when reading CSV files, which makes this option the most user friendly for this community. Additionally, it is a simple unix-level or language-level operation to filter these as a pre-processing in a robust fashion.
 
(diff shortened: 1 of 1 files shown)


balhoff:
Seems like yaml lines are denoted by a single hash at the beginning of a line. Other hashes that are part of URIs would not be interpreted in any way. I don't see an issue, unless I'm just missing it.

matentzn:
The problem is that we want this to work:
```
pd.read_csv(sssom_file, sep=\t,' , comment='#')
```

balhoff:
Oh I see, I'm not familiar with that API. I handled it in my code like this: https://github.com/NCI-Thesaurus/ncit-utils/blob/bc2a1b207b085f9abf9adbcd62ae15a6e31b86bb/src/main/scala/org/renci/ncit/ReplaceMappedTerms.scala#L57

matentzn:
The toolkit now deals with this issue by checking if a line starts with a #, otherwise ignoring it.

PR #69, opened 2021-07-16
Proposal for a 5-Star system for mappings

matentzn:
@cmungall @cthoyt @dosumis Please take a look


SSSOM.md
```diff
@@ -491,3 +492,77 @@ Not part of spec, clean up.
...
+     - `match_type`(s) (Lexical, Logical match, HumanCurated etc)
+     - `date` of the mapping
+     - `creator_id`
+     - `subject_source`
+     - `object_source`
+     - `subject_source_version`
+     - `object_source_version`
+     - `mapping_tool` if the mapping was automatically computed using a tool
+4. _4-star mappings_ fulfil all the criteria for 3-star mappings and furthermore
+   - Register there mapping at a mapping commons
```
cthoyt, line 540:
```suggestion
   - Register the mapping at a mapping commons
```

SSSOM.md
```diff
@@ -491,3 +492,77 @@ Not part of spec, clean up.
...
+   - record the following additional metadata: 
+     - `match_type`(s) (Lexical, Logical match, HumanCurated etc)
+     - `date` of the mapping
+     - `creator_id`
+     - `subject_source`
+     - `object_source`
+     - `subject_source_version`
+     - `object_source_version`
+     - `mapping_tool` if the mapping was automatically computed using a tool
+4. _4-star mappings_ fulfil all the criteria for 3-star mappings and furthermore
```
cthoyt, line 539:
```suggestion
4. _4-star mappings_ fulfill all the criteria for 3-star mappings and furthermore
```

SSSOM.md
```diff
@@ -491,3 +492,77 @@ Not part of spec, clean up.
...
+     - if the mapping is `SemanticSimilarity` (graph similarity, neighbourhood, cosine similarity), you should provide:
+       - `semantic_similarity_score`
+       - `semantic_similarity_measure`
+     - For now, if there are _multiple pieces of evidence_ (lexical, logical etc), please emit one row per evidence. 
+       If your tool combines multiple pieces of evidence in a complex way, emit yet another row at the end with
+       `match_type` `Complex` and emit ensure you provide the `mapping_tool`.
+5. _5-star mappings_ fulfil all the criteria for 4-star mappings and furthermore
+   - Are being up-to-date with the `subject_source` and `object_source`
+   - Have no issue on their issue tracker open for more than 3 months without an interaction
+   - Use a _standard_ open license, such as CC-0 or CC-BY.
```
cthoyt, line 566:
```suggestion
   - Use a _standard_ open license, such as [CC Zero 1.0](https://creativecommons.org/publicdomain/zero/1.0/) or [CC-BY-4.0](https://creativecommons.org/licenses/by/4.0/).
```

cthoyt:
Here's the link for viewing this document rendered on GitHub https://github.com/mapping-commons/SSSOM/blob/minimum-metadata/SSSOM.md#minimum @matentzn maybe you can include that in the description at the top

SSSOM.md
```diff
@@ -491,3 +492,77 @@ Not part of spec, clean up.
...
+
+Current mappings are extremely hard to use for data integration, because they are:
+
+- non-transparently imprecise
+- non-transparently incomplete
+- inaccurate
+- unFAIR
+
+In principle, to reach full integration through mappings, you will have to cross-map all ontologies, or semantic
+spaces, which means if you have N "spaces", you have N x N (-1) mappings. To mitigate the explosion
```
cthoyt, line 508: either `N choose 2` or `N x (N-1) / 2`

SSSOM.md
```diff
@@ -491,3 +492,77 @@ Not part of spec, clean up.
...
+of mappings, we have to be able to cross-walk:
+- multi-hop forward walks `{ O1:A->O2:A, O2:A->O3:A } --> {O1:A->O3:A}`
```
cthoyt, line 510: this notation is very confusing - this should be given with concrete examples. I think this could also mitigate confusion about the "walk" terminology

SSSOM.md
```diff
@@ -491,3 +492,77 @@ Not part of spec, clean up.
...
+- walk-backs `{ O1:A->O2:A } --> {O2:A->O1:A}`
+- combinations `{ O1:A->O2:A, O2:A->O3:A, O4:A->O3:A } --> {O1:A->O4:A}`
+
+To enable cross-walking, we propose the following 5 star system for mapping metadata. 
+
+1. _1-star mappings_ fulfil the following criteria:
```
cthoyt, line 516:
```suggestion
1. _1-star mappings_ fulfill the following criteria:
```

SSSOM.md
```diff
@@ -491,3 +492,77 @@ Not part of spec, clean up.
...
+   - record subject id, object id and mapping precision
+   - using qualified names (either URIs or CURIEs + curie maps) for subject id and object id
+   - using a standard file format (JSON, XML, CSV, TSV)
+   - made available in a public space
+   - _optional_: record the subject and object labels to make it easier for humans to read the file
+2. _2-star mappings_ fulfil all the criteria for 1-star mappings and furthermore
```
cthoyt, line 522:
```suggestion
2. _2-star mappings_ fulfill all the criteria for 1-star mappings and furthermore
```

SSSOM.md
```diff
@@ -491,3 +492,77 @@ Not part of spec, clean up.
...
+   - made available in a public version control system with an issue tracker
+   - record the semantic predicate explicitly
+   - using qualified names for the semantic predicate (i.e. owl:equivalentClass, skos:exactMatch)
+   - record a confidence value for the mapping between 0 and 1.
+   - record an open license for the use of the mapping set
+3. _3-star mappings_ fulfil all the criteria for 2-star mappings and furthermore
```
cthoyt, line 528:
```suggestion
3. _3-star mappings_ fulfill all the criteria for 2-star mappings and furthermore 
```

SSSOM.md
```diff
@@ -491,3 +492,77 @@ Not part of spec, clean up.
...
+     - if the mapping is `HumanCurated`.. (this needs to be [fleshed out](https://github.com/mapping-commons/SSSOM/issues/57). 
+       For now, leave a `comment` indicating what you did to arrive at your conclusion, i.e. wether you compared the definitions, 
+       looked up the "labels" in a database, ran a tool and decided to trust it etc.)
+     - if the mapping is `SemanticSimilarity` (graph similarity, neighbourhood, cosine similarity), you should provide:
+       - `semantic_similarity_score`
+       - `semantic_similarity_measure`
+     - For now, if there are _multiple pieces of evidence_ (lexical, logical etc), please emit one row per evidence. 
+       If your tool combines multiple pieces of evidence in a complex way, emit yet another row at the end with
+       `match_type` `Complex` and emit ensure you provide the `mapping_tool`.
+5. _5-star mappings_ fulfil all the criteria for 4-star mappings and furthermore
```
cthoyt, line 563:
```suggestion
5. _5-star mappings_ fulfill all the criteria for 4-star mappings and furthermore
```

SSSOM.md
```diff
@@ -491,3 +492,77 @@ Not part of spec, clean up.
...
+   - Are being up-to-date with the `subject_source` and `object_source`
```
cthoyt, line 564:
```suggestion
   - Are up-to-date with the `subject_source` and `object_source`
```

SSSOM.md
```diff
@@ -491,3 +492,77 @@ Not part of spec, clean up.
...
+     - `subject_source_version`
+     - `object_source_version`
+     - `mapping_tool` if the mapping was automatically computed using a tool
+4. _4-star mappings_ fulfil all the criteria for 3-star mappings and furthermore
+   - Register there mapping at a mapping commons
+   - record the following additional metadata:
+     - `mapping_set_id`
+     - `mapping_set_description`
+     - `mapping_set_version`
+     - `mapping_provider` (if the mapping is not original, i.e. it is not derived from another source)
```
cthoyt, line 545: the rules for what are valid mapping providers probably need to be fleshed out. Are these OBO prefixes/other stuff that might appear in the Bioregistry or FAIRSharing?

SSSOM.md
matentzn: Its not even `N x (N-1) / 2` -> because you cant actually walk back!

SSSOM.md
```diff
@@ -491,3 +492,77 @@ Not part of spec, clean up.
...
+     - `creator_id`
+     - `subject_source`
+     - `object_source`
+     - `subject_source_version`
+     - `object_source_version`
+     - `mapping_tool` if the mapping was automatically computed using a tool
+4. _4-star mappings_ fulfill all the criteria for 3-star mappings and furthermore
+   - Register the mapping at a mapping commons
+   - record the following additional metadata:
+     - `mapping_set_id`
```
cthoyt, line 542: maybe mention that this ID should be unique in combination with the `mapping_provider`

SSSOM.md
```diff
@@ -491,3 +492,77 @@ Not part of spec, clean up.
...
+- non-transparently incomplete
+- inaccurate
+- unFAIR
+
+In principle, to reach full integration through mappings, you will have to cross-map all ontologies, or semantic
+spaces, which means if you have N "spaces", you have N x N (-1) mappings. To mitigate the explosion
+of mappings, we have to be able to cross-walk:
+- multi-hop forward walks `{ O1:A->O2:A, O2:A->O3:A } --> {O1:A->O3:A}`
+- walk-backs `{ O1:A->O2:A } --> {O2:A->O1:A}`
+- combinations `{ O1:A->O2:A, O2:A->O3:A, O4:A->O3:A } --> {O1:A->O4:A}`
```
cthoyt, line 512: I think this brings up an important (so far implicit) idea about mappings - are they directed or not? Can we think of a mapping network only as a directed graph (example 1) or as an undirected graph (examples 2 and 3)?

SSSOM.md
cthoyt: If you like this example, I'll update the other two on walk-backs and combinations to match it

matentzn: Yes, this is great! Thanks for looking at this.. I am preparing the workshop now, and I will present this 5-tier system here as well.

SSSOM.md
matentzn: I think we must think of them as directed. I believe that in 99.9% of the cases you should be able to do a walk-back if you know what the inverse of the predicate used was; but we should be clear that walk-backs are still "inferences" - and therefore should be treated with slightly less confidence than their corresponding root mapping.

SSSOM.md
matentzn: I think we should actually propose this to be globally unique, using a w3id purl. I would really like that!

SSSOM.md
matentzn: I think if the mapping set is original, i.e. not derived from some source, than we would do users a disservice to overly specify the mapping_provider. I will raise this issue at the workshop - I am sure this is a super standard provenance problem and I will try to get a provenance person there as well.

matentzn:
@cthoyt thank you for all your feedback on this, operationalising the tiers is a great idea. I would suggest we propose the system first, and then implement an `sssom report` method in sssom-py which basically checks "tier adherence". See https://github.com/mapping-commons/sssom-py/issues/112

SSSOM.md
cthoyt: ORCID ID of the main responsible person? or wikidata/ROR ID for an organization if responsibility is _truly_ decentralized?

SSSOM.md
cthoyt: Agree for global uniqueness. what would such a PURL point to? Should people have to register these somewhere? If we were to provide a way to register, maybe we could end-to-end automate it via github issues like I did in the bioregistry and also have a website get autogenerated

cthoyt: I could set that up in a different repo in the mapping-commons namespace if the idea makes sense

matentzn: Yes, I already have some basic idea how this could work; the only piece missing is workflow: basically my idea is that if you run a `mapping-commons`,  you basically make a registration request on a central mapping-commons repo, or you go straight to https://github.com/perma-id/w3id.org/blob/master/sssom/.htaccess to make a PR. 

It is a bit risky; I am wondering whether w3id would accept file protection on a user basis. Alternatively, is there a way to manage rewrite rules (say for  `http://w3id.org/sssom/mappings/my-commons/my-mapping.sssom.tsv`)  somehow outside htaccess and import them? No idea.

SSSOM.md
matentzn: I think both should be permissible, yes.

cthoyt: So in general, we should ask the `mapping_provider` be a valid CURIE and give a warning if one of the preferred prefixes are not used

matentzn: What if the mapping provider is a database? it seems a bit odd then to require a CURIE rather than a URL.. This is not the same as `subject_id` imo..

cthoyt: use a [fairsharing](http://bioregistry.io/registry/fairsharing) CURIE corresponding to the database :)

You can't trust URLs don't go away over time due to databases shutting down from funding drying up/people moving on, but an entry in an external database like FAIRSharing, WikiData, or even a Bioregistry CURIE could have much more longevity

matentzn: I agree this would be better.. However, how practical for all cases I don't know.. Lets discuss it for a minute at the workshop.

matentzn:
I think I dealt with most of your feedback;

Some of your suggestions about clearer descriptions of the elements need to be done separately on the SSSOM metadata itself.

matentzn:
@cthoyt after a lot of back and forth in my head, I still don't dig the "tier" notion. I would like something that 
- implies quality: 1-star mappings are bad, 5-star mappings are good (like in [linked data](https://www.ontotext.com/knowledgehub/fundamentals/five-star-linked-open-data/)) - tier is a bit too neutral for my liking. The "good" and "less good" should really scream from the name
- that is commonly understandable - the word should mean something "publishing tier 3 mapping" means little to normal folks

I would suggest we go back over that point. Can you describe again what your problem with the star notion is, and if the problems are valid, maybe something more suitable as an alternative?

cthoyt:
@matentzn if your goal is to imply 1 is bad and 5 is good, then I think keeping the "star" nomenclature is actually a good fit. Just keep in mind that it makes it a bit more rigid - if we want to add an additional 6th level later, it will seem very strange (I expect star systems to strictly be out of 5 or 10)

matentzn:
Thank you.. I dont expect any more levels.. If anything, clarifications on the existing ones.

PR #96, opened 2021-09-29
Add build workflow

### LINKED ISSUE #95: Auto-deploy docs when sssom.yaml is changed
see https://github.com/mapping-commons/sssom/commit/4c8e1d39b4b00bce3606b51d1116370fdd8b07ef#commitcomment-57132012

we should be able to have a GH actions autodeploying the docs when sssom.yaml is changed on master

cthoyt:
Closes #95  - all that was needed was to make sure pipenv was installed then `make` did the rest :)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,29 @@
+name: Build
+on:
+  push:
+    paths:
+      - 'sssom/model/schema/sssom.yaml'
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@master
+        with:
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
```diff
@@ -0,0 +1,31 @@
...
+          python-version: 3.9
+      - name: Create local changes
+        run: |
+          pip install pipenv
+          make
+      - name: Commit files
+        run: |
+          git config --local user.email "action@github.com"
+          git config --local user.name "GitHub Action"
+          git commit -m "Update" -a
```
cthoyt, line 26: maybe change the commit message?

matentzn (approved):

PR #99, opened 2021-10-26
Add predicate modifier / change entity range to EntityReference over uriorcurie

### LINKED ISSUE #40: Negated mappings and the standardisation of mapping predicate modifiers
This issue is a history of the discussion on how to handle negated mappings. After a lot of discussion and a final vote at https://github.com/mapping-commons/sssom/issues/40#issuecomment-942298272, we've decided to go with adding an additional predicate modifier column to the SSSOM standard. This issue can be closed along with a pull request that realizes this update.

See draft solution in #99

Original issue text from @cmungall:
> Similar to #38 we could allow predicates to be property expressions such as `!owl:equivalentTo`

cthoyt:
Closes #40 

Following the vote to add a predicate modifier to denote negated mappings, this PR is a stub for the realization by updating the schema. @matentzn feel free to finish this one the way that you think looks best


model/schema/sssom.yaml
```diff
@@ -22,6 +22,11 @@ default_curi_maps:
 default_prefix: sssom
 
 enums:
+  predicate_modifier_enum:
+    permissible_values:
+      Not: Negating the mapping predicate. The meaning of the triple becomes subject_id is not a predicate_id match to object_id.
+      Direct: States that the mapping predicate directly relates two entities connected by a transitive predicate like rdfs:subClassOf, i.e. there is no other entities in between.
+      Inverse: Inverting the predicate. The meaning of the triple becomes subject_id is inverse-related with predicate_id match to object_id (effectively turning the direction around).
```
cthoyt, line 29: I think we explicitly agreed not to add direct/inverse as predicate modifiers

cthoyt: or at least to table that discussion for later

matentzn: So we only add "Not" at the moment? I thought @cmungall main use case were Inverse and Direct.

matentzn: I removed Indirect and Inverse for now - we can add them later. One evil at a time.

PR #103, opened 2021-11-03
Applying minimal entity modelling

matentzn:
See also https://github.com/linkml/linkml/issues/450

The question is whether (and which) element in SSSOM should be modelled as nodes, and which as uriorcurie.

@cmungall is this the right modelling syntax?

The question that remains is what to do about all the other fields, say `creator_id`. Shall we simply switch all these to datatype `Uri`, to avoid having to deal with the ambiguity of the field (without any magic code, uriorcurie would be translated to "orcid:123"^^xsd:anyUri, rather than "http://orcid.org/123"^^xsd:anyUri when "orcid:123" is supplied in the sssom table).

### DIFF
--- a/model/schema/sssom.yaml
+++ b/model/schema/sssom.yaml
@@ -65,9 +65,12 @@ slots:
     description: Contains a list of mapping objects
     range: mapping
     multivalued: true
+  id:
+    description: CURIE or IRI identifier
+    identifier: true
   subject_id:
     description: The ID of the subject of the mapping.
-    range: uriorcurie
+    range: entity
     required: true
     mappings:
     - owl:annotatedSource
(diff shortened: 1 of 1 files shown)


cmungall (approved):
This is correct.

Note this doesn't impact the json serialization, since we are not inlining it. Similarly it should not impact the TSV serialization.

The python OM interface becomes a little different - instantiating should be the same, but accessing will require an extra level of indirection (`mapping.subject_id.id`).

It also complicates the simplicity of the model and invites questions such as why doesn't the entity class have a label?

cmungall:
While this approach is not wrong, we want to avoid the overhead of having a stub Entity class

See
https://github.com/linkml/linkml/issues/450

Upgrade linkml-runtime

Create a new type, as per the example there e.g.

```
types:
 IdReference:
    typeof: uriorcurie
    description: A reference to a mapped entity. This is represented internally as a string, and as a resource in RDF
    base: str
    uri: rdfs:Resource
```

and use this as your range for subject_id, predicate_id, object_id

matentzn:
Closed in favour of #99 which has the biolink preferred solution for this

PR #109, opened 2021-12-01
added inlined_as_list for mappings and regenerated

hrshdhgd:
This is a change made with reference to the [linkml issue ](https://github.com/linkml/linkml/issues/477) and the [sssom-py issue](https://github.com/mapping-commons/sssom-py/issues/177)


matentzn (approved):
@hrshdhgd this is good, but for the future, @cthoyt has provided a system to automatically update derived data when the schema is updated.

PR #110, opened 2021-12-02
Making some fields recommended

cmungall:
https://w3id.org/linkml/recommended

has no effect on validation, json-schema generation, or python generation

tools/applications can optionally use `recommended` to issue warnings

It's not clear if we can come up with universal guidelines for what is required, in future we may want to define profiles. For now I have added subject and object label since this makes the mappings more approachable to human users.


matentzn:
Note also 

https://mapping-commons.github.io/sssom/spec/#the-sssom-5-star-system-for-mappings

PR #114, opened 2021-12-02
link to spec

cmungall:
the existing link is to a soft redirect

or use w3id?


cthoyt:
I'd say go right for the w3id

PR #117, opened 2021-12-03
Making match_type recommended

matentzn:
At the moment, 4 elements in sssom are required, i.e. validation fails if they are not present:

- `subject_id`
- `object_id`
- `predicate_id`
- `match_type`

This has big ramifications for the toolkits, which will simply fail if one of the four is not present.

I hereby move forward the motion to make `match_type` `recommended` instead of `required`. This is not to diminish its importance, it's just impractical to have all our toolchains fail hard when processing simple <s,p,o> tables.

This PR will be merged on Friday 10th December unless objections.

### DIFF
--- a/model/schema/sssom.yaml
+++ b/model/schema/sssom.yaml
@@ -165,7 +165,7 @@ slots:
   match_type:
     description: The kind of match that led to the mapping, e.g. Logical or Lexical.
     range: match_type_enum
-    required: true
+    recommended: true
     multivalued: true
     examples:
       - value: Lexical


bpeters42:
Sorry if I missed this discussion earlier, but would it be possible to push users more towards filling this out rather than allowing it to be blank? If making it required is too heavy handed, maybe set it to a default value when left blank, and make that default value 'unspecified match type', or something else that makes the people creating the mapping realize that they could be more specific.

matentzn:
Good point @bpeters42 

@cmungall is this a case for default values? Can we just add on ("Unspecified")? Is this better then making it recommended?

cthoyt:
I object, I think this is a reasonable amount of metadata to require. All SSSOM tables had to be constructed at some point - if they're sucked of an ontology, then you can mark it as manually curated, etc. A better alternative than making the schema more permissive towards ambiguous, low quality mappings would be better documentation on on what people should put in common situations

graybeal:
The toolchains that will fail aren't built yet, right? (Not being facetious, just not sure what you're referring to.)  

Is the concern with requiring this that it is hard to provide, or unknowable, or maybe just kind of annoyingly redundant? The last could be fixed by having a match_type for the entire set of matches. Unknowable is addressed by 'unknown'. For me, the one concern that justifies not requiring this is that the list of match_types is not rigorously defined.

Well, and also note the 5-star system only requires match_type for 3 stars and above.

(I'm using https://mapping-commons.github.io/sssom as my reference, I trust that's the right place.)

Two definitions (list is below) are recursive and somewhat ambiguous (what's logical to me may not be logical to you, ha ha); there is possibility that multiple types apply (it can be lexical, human-curated, and complex, for example; whether/how to blend types is not discussed); I don't know what Semantic Similarity means (I assumed it detected this through logical chains, but now I'm thinking it's redundant.

And the definition in the glossary is equally unhelpful: The kind of match that led to the mapping, e.g. Logical or Lexical.`

So I'd say if you're going to make it required, and maybe even if not, it needs to be much better defined, so that it will be able to serve the intended purpose. 

Lexical | Lexical match |   |  
Logical | Logical match |   |  
HumanCurated | Match based on human expert opinion |   |  
Complex | Match based on a variety of different strategies |   |  
Unspecified | Unknown match type |   |  
SemanticSimilarity | Match based on close semantic similarity

matentzn:
We will continue the discussion on this soon, and all the issues raised here will be addressed. Thank you both

matentzn:
I have an entirely different suggestion to move forward here, so will close this.

cmungall:
I'm sorry I didn't spot this the first time around. I think making this required inflates file sizes and communicates zero bits of information. If we think that having an enum value Unspecified carries more information than a null value then we are fooling ourselves into thinking we are more rigorous than we are. This is just inflating file sizes and incurring non-zero I/O latency costs for no reason.

Issue #116, opened 2021-12-03
Unbreak autopushing docs


matentzn:
https://github.com/mapping-commons/sssom/runs/4407903105?check_suite_focus=true

For some reason build.yml does not work anymore - @hrshdhgd do you mind looking at that for a moment?

matentzn:
Maybe while you are at it, combine with https://github.com/mapping-commons/sssom/issues/115?

hrshdhgd:
The build.yml works as expected now and highlighted the issues along the way.

1. The `Pipfile` and the `Pipfile.lock` are currently present in the `make-venv` directory. 
    - Because of this, pipenv cannot see these two files when running within GitHub actions (locally as well for me) and hence creating new files (with none of the necessary packages listed). Thus the test was failing at `Module not found: linkml_runtime`
    - **Solution**: I copied the two files to the home directory and moved past this error to the next one
    - **Question**: Should we just move the 2 files to the home directory as opposed to having them within 'make-venv' directory?

2. `ImportError: cannot import name 'Person' from 'sssom.sssom' (/home/runner/work/sssom/sssom/sssom/sssom.py)`
    - This error arises from [this line](https://github.com/mapping-commons/sssom/blob/b48d18cef70130628639e74bea42032a435ae7ce/tests/test_input_against_model.py#L6). There is no definition of the `Person` class in the project. How do I proceed?

cc @matentzn , @cmungall

matentzn:
@hrshdhgd has fixed this.

Issue #115, opened 2021-12-02
Add pr-check.yaml to actions


cmungall:
We have one action to build products #95. 

We should also have one that runs unit tests, should just be `make test`

Example
https://github.com/linkml/linkml-model/blob/main/.github/workflows/pr-test.yaml

Also consider:
https://github.com/linkml/linkml-model/blob/main/.github/workflows/pypi-publish.yaml

hrshdhgd:
Is this still an outstanding issue? If so, what needs to be done? All PRs linked to this thread seem merged or closed.

matentzn:
Right now we have two tests:

1. https://github.com/mapping-commons/sssom/blob/9c2af819b212825a89d3dbd5c43f762ba983d065/Makefile#L54
2. https://github.com/mapping-commons/sssom/blob/9c2af819b212825a89d3dbd5c43f762ba983d065/project.Makefile#L41

Neither of these are comprehensive - the first basically makes sure the schema can be build, the second checks on a small set of test cases in the sssom repo if they are still valid or not - based on the latest version on PyPI rather than the version on the `main` branch.

To finish this issue, we need to have a simply way to test the schema when changes are proposed. So basically, we need to:

1. Build the data classes
2. Somehow check if with the new data classes, a set of test cases can still be passed

I don't know @cmungall full vision on this. As we cannot use linkml convert to parse the TSV files, we may have to build a test suite of 5-10 mapping sets in JSON, and see if we can use `linkml-convert` to parse them after the changes to the datamodel are proposed. It is worth though to raise this with him and see how he would suggest this to be done, as I assume this should be the exact same thing for all linkml projects.

PR #118, opened 2021-12-03
Issue 155 and 116

hrshdhgd:
- [ ] Get `make test` under GitHub Actions (#115 )
- [ ] Fix the error in #116


.github/workflows/build.yml
```diff
@@ -24,8 +26,12 @@ jobs:
...
-          github_token: ${{ secrets.GITHUB_TOKEN }}
-          branch: ${{ github.ref }}
+      # - name: Push changes
+      #   uses: ad-m/github-push-action@master
+      #   with:
+      #     github_token: ${{ secrets.GITHUB_TOKEN }}
+      #     branch: ${{ github.ref }}
+      - name: Run unit tests
+        run: |
+          make test
```
cthoyt, line 36: what happens when the tests fail? does it make this command fail, and therefore the action fail?

cthoyt:
not sure why github is automatically merging commits with failing builds, this seems like a really bad idea

hrshdhgd:
This is horrible ... I am not sure what triggered this

PR #140, opened 2022-02-11
Include OWL files. Possibly fixes #102

### LINKED ISSUE #102: Review JSON-LD and RDF serialisations
Once the latest big [push on sssom-py](https://github.com/mapping-commons/sssom-py/pull/172) is completed, we should review the RDF and JSONLD serialisations with @graybeal and other people in the FAIR/Semantics space, and get their feedback. Who else other than John wants to be part of this? We will probably organise a call to do that.

cmungall:


matentzn (approved):

matentzn:
Hm, this is a very complex request, as all of these files are derived files using a framework entirely independent from SSSOM - (https://linkml.io/). If you wanted to articulate the need of separating imported entities more cleanly from internal ones, you should do it on their repo.

However, I doubt you will have much success - we have been pushing front and center to get rid of owl:imports, due to its weaknesses when it comes to tooling (network dependencies, special tools needed to load) and versioning (our imports change frequently, and it has become impossible to figure out exactly how "an ontology was at timepoint x". As this is very far beyond this issue here, you can, if you want, see this discussion here for further details on that matter: 

https://github.com/INCATools/ontology-development-kit/issues/194

afk314:
OK, I understand.  I suppose I haven't thought deeply enough about the topic to outright disagree with that!  They are a problem to manage but I don't rely on dereferenced ontologies ever and I guess just accept all the pain.  Thanks for the answer.

PR #153, opened 2022-04-11
Replacing `match_term_type` by `subject` and `object_type`

### LINKED ISSUE #143: Tear apart the `match_term_type` into to new elements
Currently, we have the `match_term_type`, to indicate wether a match is between classes, properties etc.

- TermMatch: A match between two terms
- ConceptMatch: A match between two SKOS concepts
- ClassMatch: A match between two OWL/RDFS classes
- ObjectPropertyMatch: A match between two OWL object properties
- IndividualMatch: A match between two OWL Individuals
- DataPropertyMatch: A match between two OWL object properties

This is a bit limited because we may have to map between different kinds of elements. Before finalising version 1, we would like to propose to remove this field in favour of a more fine grained approach that allows to provide typing for both the subject and the subject, like:

- `subject_type` (owl:Class, rdf:Resource, skos:Concept etc etc)
- `object_type`
- `predicate_type` (owl:AnnotationProperty)

This gives us a range of advantages:
1. We can use the semantic web standard vocabularies for entity types and do not have to define our own
2. We can use the typing information in tools which is very useful for example when generating OWL mappings
3. We avoid combinatorial explosions of match_types (ObjectPropertyToAnnotationPropertyMatch, ClassToDatabaseEntityMatch)

@ernestojimenezruiz since you were the one proposing this element originally, can you provide use with some feedback on this proposal?

matentzn:
Fixes #143

This removes the `match_term_type` and its corresponding   enums, replacing it by two seperate (subject and object match types)

### DIFF
--- a/model/schema/sssom.yaml
+++ b/model/schema/sssom.yaml
@@ -413,7 +413,6 @@ classes:
     - object_source
     - object_source_version
     - mapping_provider
-    - mapping_cardinality
     - mapping_tool
     - mapping_date
     - subject_match_field
@@ -449,6 +448,7 @@ classes:
     - object_source
     - object_source_version
     - mapping_provider
+    - mapping_cardinality
     - mapping_tool
     - mapping_tool_version
     - mapping_date
--- a/model/schema/sssom.yaml
+++ b/model/schema/sssom.yaml
@@ -22,6 +22,30 @@ default_curi_maps:
 default_prefix: sssom
 
 enums:
+  entity_type_enum:
+    permissible_values:
+      owl class:
+        meaning: owl:Class
+      owl object property:
+        meaning: owl:ObjectProperty
+      owl data property:
+        meaning: owl:DataProperty
+      owl annotation property:
+        meaning: owl:AnnotationProperty
+      owl named individual:
+        meaning: owl:NamedIndividual
+      skos concept:
+        meaning: skos:Concept
+      rdf resource:
+        meaning: rdf:Resource
+      rdf class:
+        meaning: rdfs:Class
+      rdf literal:
+        meaning: rdfs:Literal
+      rdf datatype:
+        meaning: rdfs:Datatype
+      rdf property:
+        meaning: rdf:Property
   predicate_modifier_enum:
     permissible_values:
       Not: Negating the mapping predicate. The meaning of the triple becomes subject_id is not a predicate_id match to object_id.
@@ -87,7 +111,7 @@ slots:
         description: (The CURIE of the biolink class for genes.)
   subject_type:
     description: The type of entity that is being mapped.
-    range: EntityReference
+    range: entity_type_enum
     examples:
       - value: owl:Class
   predicate_id:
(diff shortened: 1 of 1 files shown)


model/schema/sssom.yaml
```diff
@@ -93,6 +85,11 @@ slots:
         description: (The CURIE of the Uberon term for "anatomical entity".)
       - value: biolink:Gene
         description: (The CURIE of the biolink class for genes.)
+  subject_type:
```
cthoyt, line 88: name could be confused with the entity type (like disease or protein)

matentzn: I thought of that, but I guess this would be the case with these kinds of words no matter what. It was previously "type" (the state before this change). Any idea for an alternative? I kinda thought: rdf:type, so... `type`. At least some people wont be confused :)

cthoyt: Can you please provide an example row in an SSSOM TSV that would use these entries in the PR description?

cmungall: Even in rdf terms it could be confusing. I am both rdf type namedindividual and Homo sapiens. I think the name is ok if sufficiently documented though, and besides. An alternative is owl_type

cmungall (approved):

model/schema/sssom.yaml
cthoyt: so the goal here is to put verrry high level entity types? If that's the case, can we have an enumeration of a very small set of valid ones? or is this just a place where we expect people to put a slurry of things and then later we have to write tools to deconvolute it?

matentzn: @cmungall according to @cthoyt I restricted the set of legal `types`. What is your opinion on this?

matentzn: (also on _how_ I did it)

cthoyt: Thanks Nico, this is now much easier to understand. Entity type could be inferred from external resources that aren't SSSOM and this could be considered scope creep, but also that's not always so easy to do, so having it explicit could be nice. At least I think the proposed update is better than what was there before.

matentzn: Exactly. This is purely optional metadata, which can be used by sssom-py for certain assumptions, like how to render an "OWL" or a "SKOS" version of the mapping. We do not always know the source, or have a standard mechanism to determine just based on the subject ID.

cthoyt (approved):

model/schema/sssom.yaml
cmungall: I think this is a good and practical set of enums

If we are doing this I think we may need to do some work on specifying the owl mapping

E.g. how do I interpret A subClassOf B in the following scenarios:

- A is punned (subject_type: [Class, Individual]
- A is an individual only
- A is an individual B is a class
- A and B are a mix of OWL and SKOS types

These are pretty obscure edge cases but nevertheless we should document

cmungall (approved):
I think this makes sense

matentzn:
Thank you @cthoyt and @cmungall

Issue #150, opened 2022-03-25
Breaking change: mapping justification, curation rules


matentzn:
Before moving to SSSOM version 1.0. I would like to suggest one major breaking change for SSSOM. This is coming out of our growing experience how mapping sets are being used moving forward: Let's get this right this time!

**Summary**:

- change `match_type` to `mapping_justification` ([issue](https://github.com/mapping-commons/sssom/issues/10)). **A mapping justification is an action** (or the written representation of that action) **of showing a mapping to be right or reasonable**. This is to:
   - get rid of the confusion `match/mapping` dichotomy (thanks @graybeal to an excellent [discussion](https://github.com/mapping-commons/sssom/issues/10) clarifying this)
   - solidify the core data model of SSSOM to be a mapping (subject, predicate, object) plus a justification for the mapping, which can be anything from a human curator decision to a tool match
   - Other groups have been using mapping justifications all along #85
- move `mapping_justification` from an enum (`HumanCurated`) to a proper controlled vocabulary (`vocab:HumanCurated`) #141 
   - There will be many more justifications than the few we have already covered, and it is easier to manage these in a controlled vocabulary than an enum of a datamodel.
- Explicitly introduce the notion of `curation_rule` as a separate element that applies (mostly or exclusively) to Human-curated mappings. While closely related, a mapping justification would, as @AlasdairGray describes [here](https://github.com/mapping-commons/sssom/issues/85#issuecomment-914104106) be the `application of the curation rule in a specific instance with details of the parameters that were used`.  See [here](https://github.com/mapping-commons/disease-mappings/issues/16) for an example discussion on mapping rules in the area of disease and phenotypes.

I expect a lengthy discussion here so feel free to weigh in!

matentzn:
Closing this as the only thing that remains to be done is https://github.com/mapping-commons/sssom/issues/166

PR #154, opened 2022-04-11
Breaking change: Replacing `match_type` with `mapping_justification`

matentzn:
See also #150

After some deliberations with some of you, this will be the hopefully last major breaking change in SSSOM.

- [ ] Note to self, make comment on https://github.com/agroportal/project-management/issues/265 after PR is merged

### DIFF
--- a/model/schema/sssom.yaml
+++ b/model/schema/sssom.yaml
@@ -159,10 +159,8 @@ slots:
     range: EntityReference
     required: true
     examples:
-      - value: SSSOMV:Lexical
-        description: (The type of match that led to the mapping is Lexical, i.e. based on comparing the string representation of an aspect of the subject with the string representation of an aspect of the subject.)
-      - value: SSSOMV:HumanCurated
-        description: (The type of match that led to the mapping was by human expert judgement.)
+      - value: OMAPV:Lexical
+      - value: OMAPV:HumanCurated
   mapping_set_id:
     description: A globally unique identifier for the mapping set (not each individual
       mapping). Should be IRI, ideally resolvable.
--- a/model/schema/sssom.yaml
+++ b/model/schema/sssom.yaml
@@ -33,14 +33,6 @@ enums:
       "1:0": One-to-none mapping
       "0:1": None-to-one mapping
       "n:n": Many-to-many mapping
-  match_type_enum:
-    permissible_values:
-      Lexical: Lexical match
-      Logical: Logical match
-      HumanCurated: Match based on human expert opinion
-      Complex: Match based on a variety of different strategies
-      Unspecified: Unknown match type
-      SemanticSimilarity: Match based on close semantic similarity
   match_term_type_enum:
     permissible_values:
       TermMatch: A match between two terms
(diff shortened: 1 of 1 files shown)


model/schema/sssom.yaml
```diff
@@ -162,15 +154,14 @@ slots:
...
-    description: The kind of match that led to the mapping, e.g. Logical or Lexical.
-    range: match_type_enum
+  mapping_justification:
+    description: A mapping justification is an action (or the written representation of that action) of showing a mapping to be right or reasonable.
+    range: EntityReference
     required: true
-    multivalued: true
     examples:
-      - value: Lexical
+      - value: SSSOMV:Lexical
```
cthoyt, line 162: Before evaluating if this is a good change for SSSOM, I think the `SSSOMV` vocabulary needs to get fleshed out a bit more? I.e., is there finally a page that is specifically describing all of the terms? Previously, there were some explanations of a few of the possible terms buried in a pile of other irrelevant documentation, which isn't good enough.

This vocabulary also needs a Bioregistry entry ;)

matentzn: Maybe I should remove the example - This vocabulary does not exist yet.. Not even sure its should have the SSSOM namespace, or another: Something like "Vocabulary for Ontology Mapping"? That could be a cool standalone thing to do anyways, and then the prefix wouldn't be SSSOMV. What do you think? A vocabulary in the SSSOM namespace or adjacent (SSSOMV) or something independent in mapping commons org: Vocabulary for Ontology Matching?

graybeal: I too am having trouble evaluating this change. Partly understanding/tracking the purpose of the diverse changes, partly understanding their applications (what does it mean when permission_values are or are not listed?), partly interpreting the `value:` string generally (what does it mean that most do not have full IRIs or CURIEs? are they not in the vocabulary)?  Some of the definitions, n.b. Logical and Lexical, are inadequate, and some of the choices are overlapping.

My observation on that particular point is that the vocabulary name/IRI should be defined as part of the release. Include SSSOMV in the prefix list and make it `https://w3id.org/sssom/schema/voc/` (please note the https protocol, which SSSOM should also be define as). You already have the w3id prefix so this should not be an obstacle to implementation at any point.

cthoyt:
@matentzn btw it's already sort of a problem that the O in SSSOM stands for ontologies since there are so many other kinds of semantic space formats. if you want to mint a new vocabulary to describe mappings, I'd suggest not making it specifically named for ontologies

model/schema/sssom.yaml
matentzn: For the sake of explaining the situation, I created the vocabulary I envision here, with only 3 classes: https://github.com/mapping-commons/ontological-mapping-vocabulary. I propose to build a separate vocabulary, independent of SSSOM, to capture all the activities, match types etc as they are currently prevalent in the mapping community. I know about all the disagreements on terminology, so let us not make that an issue here - this is already in SSSOM - so not currently up for debate.

So this PR should not be about evaluating that vocabulary - I will complete it with what we anyways have right now in SSSOM once the general approach is signed off on. And then anyone can add new terms in there.

The only thing that matters here is this:

- We change the `match_type` field to `mapping_justification` for the reasons outlined in #150 
- We constrain the `mapping_justification` field to be values from a controlled semantic web vocabulary.

matentzn:
@cthoyt Name suggestions welcome :P

cmungall (approved):

matentzn:
@callahantiff @graybeal can you also take another look?

graybeal (approved):

matentzn:
Thank you all for the review!

callahantiff:
So sorry I was unable to review in time @matentzn. I will make sure to respond quicker in the future!

matentzn:
NP I would have waited, am just in a rush on some features :)

callahantiff:
> NP I would have waited, am just in a rush on some features :)

I completely understand! 😄

PR #156, opened 2022-04-15
Temporary fix for validating mapping justifications

matentzn:
At the moment, we cannot automatically validate slots/columns against external vocabularies due to limitations in the LinkML validation framework. To ensure that we have some kind of control in the meantime, this PR re-introduces the mapping justifications as enums with the enums being mapped to an external vocabulary. Ultimately, this enum should simply go away, with no breaking changes being introduced as a consequence.

Am I understanding correctly that if I use `meaning` in conjunction with `slots` that I can pass the EntityReferences in meaning (e.g. omapv:LexicalMatching) to the object model (rather than the enum ID, e.g. `lexical matching`), @cmungall?


matentzn:
Superceded by https://github.com/mapping-commons/sssom/pull/164

PR #159, opened 2022-04-15
Adding slots for mapping commons work

matentzn:


src/linkml/sssom.yaml
```diff
@@ -70,6 +70,37 @@ types:
     uri: rdfs:Resource
 
 slots:
+  mirror_from:
+    description: A list of mapping set references.
```
cmungall, line 74: Different slot?

matentzn: Sorry, fixed. Thx

src/linkml/sssom.yaml
```diff
@@ -479,3 +513,20 @@ classes:
     - other
     - comment
     class_uri: owl:Axiom
+  mapping registry:
+    description: A registry for managing mapping sets. It holds a set of mapping set references.
+    slots:
+      - mapping_registry_id
```
matentzn, line 519: declare as identifier in slot declaration.

PR #164, opened 2022-04-16
Restrict justifications to specific set

matentzn:
This PR seeks to restrict `mapping_justification` to a specific controlled vocabulary. 

However I am still unsure if this is correct. Datamodel-wise the values of `mapping_justifcations` are entity references. This solution here looks like they are validated as string? What if I change prefix? What if the entity is in URI form, i.e. I am parsing from RDF?


matentzn:
I will leave this open for the rest of this week to get people to give feedback on the `semapv`/`omapv` idea. @cthoyt @cmungall @graybeal

matentzn:
I will leave this open for the rest of this week to get people to give feedback on the `semapv`/`omapv` idea. @cthoyt @cmungall @graybeal

cmungall (approved):

cmungall (approved):

PR #173, opened 2022-05-10
Documentation Deploy

hrshdhgd:


.github/workflows/deploy_documentation.yml
```diff
@@ -2,90 +2,83 @@ name: Deploy Documentation
...
-      - name: Check out repository
-        uses: actions/checkout@v2
-        with:
-          persist-credentials: false # otherwise, the token used is the GITHUB_TOKEN, instead of your personal token
-          fetch-depth: 0 # otherwise, you will failed to push refs to dest repo
+    #----------------------------------------------
+    #       check-out repo and set-up python
+    #----------------------------------------------
+    - name: Check out repository
+      uses: actions/checkout@master
```
matentzn, line 25: is it a good idea, to use `master` on any action? What if master breaks?

matentzn (commented):
Looks great!

.github/workflows/deploy_documentation.yml
hrshdhgd: If incoming code from a PR breaks code in `master`, wouldn't that be caught in the PR itself? (assuming the PR is ahead of the master in terms of commits).

matentzn (approved):

PR #178, opened 2022-05-20
Changing preprocessing fields to controlled vocabulary

### LINKED ISSUE #142: Move `subject_preprocessing` etc from enum to controlled vocabulary
Similar to 
https://github.com/mapping-commons/sssom/issues/141

we want to move `preprocessing` fields to a controlled vocabulary. Currently, only `Stemming` is a defined preprocessing step, but we can expect to add dozens more once we start adding this kinds of metadata to outputs of tools.

matentzn:
Fixes #142


graybeal (approved):

cmungall (approved):
we could possibly revisit with linkml 1.3. which allows for dynamic enums, but this is the best way to keep it open for now

PR #182, opened 2022-05-23
Make project ready for PyPI release

hrshdhgd:
- [x] Move `sssom.yaml` to `src/sssom/datamodel/`
 - [ ] `poetry build`
 - [ ] `poetry publish`
 - [ ] Deploy to PyPI as `sssom-schema`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -132,3 +132,4 @@ dmypy.json
 .pyre/
 .DS_Store
 tmp/
+src/sssom/datamodel/sssom.yaml
--- a/Makefile
+++ b/Makefile
@@ -41,7 +41,7 @@ install:
 .PHONY: install
 
 all: gen-project gendoc
-%.yaml: gen-project
+%.yaml: gen-project move-yaml
 deploy: all mkd-gh-deploy
 
 # generates all project files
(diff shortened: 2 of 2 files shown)


hrshdhgd:
I tested this by doing the following:

- ran `make clean`
- ran `make all` 

Worked with no issues

 - ran `make build` that generates the `.whl` and `.gz` files
 - Installed the `.whl` locally in an independent fresh virtual environment and ran the following code

```
from linkml_runtime.utils.introspection import package_schemaview
import sssom_schema

view = package_schemaview('sssom_schema')
```

`view` contains contents of `sssom_schema.yaml`. This can be used to validate models in other projects .

matentzn:
> Well, the renaming was mostly done so that poetry packages the project with the name we want to deploy it in pypi. That was the main reason for the name change.

Why does poetry care about folder names? Isnt the package name defined in some config?

hrshdhgd:
> > Well, the renaming was mostly done so that poetry packages the project with the name we want to deploy it in pypi. That was the main reason for the name change.
> 
> Why does poetry care about folder names? Isnt the package name defined in some config?

Just changing it in the config apparently doesn't cut it. It goes looking for directories with the same name, else throws an error about its absence.

matentzn:
This is so lame :D But ok. Just make sure that the general `make` for gen project etc still works before merging :)

hrshdhgd:
Hahah I agree, but it is what it is. Yes, `make all` works as expected, I triple checked it.

PR #192, opened 2022-06-03
deploy using gh action

hrshdhgd:
- [x] Workflow for auto-release to PyPI


matentzn (approved):
Did you deploy that token?

hrshdhgd:
> Did you deploy that token?

Yes it is in PyPI and the repo

Issue #227, opened 2022-08-31
switch from default Jinja2 mode to explicit template mode and display examples for slots


turbomam:
I'll do this

matentzn:
This has been done - you can now make suggestions directly to the doc templates: https://github.com/mapping-commons/sssom/tree/master/src/doc-templates

Issue #226, opened 2022-08-31
Examples of `object_preprocessing` and `subject_preprocessing`


ialarmedalien:
It would be useful to provide examples of what should go in these fields as it's not clear from reading the spec and none of the SSSOM files I looked through used these fields.

matentzn:
This merrits a better answer, but just so you have something to work with while I extend the documentation, check:

https://github.com/mapping-commons/sssom/blob/master/src/sssom_schema/schema/sssom_schema.yaml#L396

And 

`matching process pre-processing` in https://mapping-commons.github.io/semantic-mapping-vocabulary/

This is not very well fleshed out yet, but feel free add additional preprocessing steps to 

https://github.com/mapping-commons/semantic-mapping-vocabulary/blob/main/semapv-terms.tsv

PR #228, opened 2022-08-31
explicit docs template dir

turbomam:
Partial solution for [#227, #226]

- [x] new doc generation rule in `project.Makefile` that uses an explicit `--template-directory`
- [x] `doc_templates/slot.md.jinja2` edited so that slot example values are displayed


turbomam:
Sorry, I committed the autogenerated `project/` files

matentzn:
Thank you @turbomam!

The jinja 2 files - shouldnt they be supplied by linkml? what happens if these need to be updated?

Issue #256, opened 2023-02-10
how to use subject_category and object_category?


saubin78:
Hi,
the documentation for both those metadata elements lacks clarity :

> The conceptual category to which the subject belongs to. This can be a string denoting the category or a term from a controlled vocabulary.

what do you mean by category? Is it a semantic one, like "disease", "material", "activity", etc.? If not, what? 
What's their purpose?

matentzn:
@saubin78 Thank you, as always, for your awesome feedback. Can you check if this is making it clearer: https://github.com/mapping-commons/sssom/pull/257

Please be strict, this should really be improved significantly.

matentzn:
@saubin78 did you get a chance to look at #257 ? Is this ok for now?

matentzn:
Fixed an release

PR #257, opened 2023-02-10
Improve documentation for subject and object categories

matentzn:
Solves #256

Resolves [#ISSUE, #ISSUE]


saubin78:
Thanks, this is much clearer with the category examples in particular. 
Yet, there is still something confusing. As you state that this "is documentation for human reviewers", I suggest that the final example be a term label, e.g. "anatomical entity", instead of an identifier "UBERON:0001062".

matentzn:
@saubin78 thats right, can you check again?

saubin78:
This looks like a good compromise as it allows using a controlled vocabulary or not, a label or a CURIE. 👍

matentzn:
Thank you @saubin78

PR #259, opened 2023-02-16
Add issue tracker item to SSSOM datamodel

### LINKED ISSUE #78: [New metadata element]: issue_tracker
(Must be lower case and contain only letters and underscores.)

```
issue_tracker
```

```
URL
```

**Description**
(Provide a human-readable description that clarifies the intended use of the metadata element.)

The issue tracker that can be used to report bugs and request changes for this mapping set.

EDIT: 

I propose to add `issue_tracker` and `issue_tracker_item` separately now; see #259

matentzn:
Fixes #78 

- [x] [CHANGELOG.md](https://github.com/mapping-commons/sssom/blob/0.9.0/CHANGELOG.md) has been updated.


sierra-moxon:
@matentzn  - do you define the type "uri" in your schema outside of linkml_types?  I don't think we have plain "uri" as a type.  You can define it more rigorously as just a "uri" in your model (with all the regex you want to restrict that).

sbello:
Is the intent that this would be a new field in the file where you could attach an issue URL to an individual mapping (e.g. a single row in the tsv file) or is this intended to be a new item in the header to point to the location of an issue tracker for the mapping project? I could see where either would be useful but it wasn't clear to me from the description. 

I would also hope that if the field is for each mapping that this field is optional since we don't usually have issues for each mapping.

matentzn:
@sierra-moxon thank you for your feedback, very much appreciated - what is the correct linkml type for URLs (rather than resources)?

@sbello thank you very much for your feedback. No SSSOM field aside subject, predicate, object and justification will ever be required. This is totally optional!

gaurav:
Some other options:
- `tracker`
- `tracked_at`
- `issue_tracker`

And, in a pinch, `seeAlso`.

sierra-moxon:
W/re to the correct type for URL (not resource), I think right now I'd use `string` if I had to choose an existing LinkML type.  Here's an example in biolink of making your own type: https://github.com/biolink/biolink-model/blob/9570188f87740e8751e05ca298abeb6ea35203af/biolink-model.yaml#L210 (none of which have a regex to define them and I think we should probably remove some of them as they are not providing any additional information to the model that `range: string` would not).  You could add a `pattern` annotation to the SSSOM-specific type designation to constrain the values of that type further than just any `string`

matentzn:
Thank you @sierra-moxon, @sbello  and @gaurav 

I think we should opt for the more verbose "issue_tracker_item", just because the issue tracker URLs are anyways super long, so a long element name wont really make much difference, and I kind of like the fact that it's not too ambiguous.

If you think I am wrong, let me know!

@sierra-moxon thanks for sharing the type spec. Do I see this correctly then that `uri` as a type is fine (if I import the linkml datatypes, which I do)?

src/sssom_schema/schema/sssom_schema.yaml
```diff
@@ -435,6 +435,18 @@ slots:
     examples:
       - value: https://www.wikidata.org/wiki/Q865360
         description: (the Wikidata identifier for the Jaccard index measure).
+  issue_tracker_item:
+    description: A URL location of the issue tracker item discussing this mapping or mapping set.
+    range: uri
```
gouttegd, line 440: Given that you have expressed concerns about issue tracker URLs always being super long, how about allowing to ”CURIfy” them?

That is:

```
#prefix_map:
#  GITHUB: https://github.com/mapping-commons/sssom/issues/
subject_id  predicate_id     object_id  issue_tracker_item
EXA:0001    skos:exactMatch  EXB:0001   GITHUB:166
```

The only problem I foresee with that is if mappings in a mapping set are tracked on more than one repository, because the repository name is part of the URL prefix (so one would need a different prefix for each repository – or choose not to use CURIfication at all and always use full URLs).

But for the (presumably) common case where there is only one issue tracker, that should work just fine.

matentzn: I thought about this as well, but I dont think it its in the spirit of CURIEs to use them like this. It would be:

GITHUB:mapping-commons/sssom/issues/166, which does not really save us much.. Doing something like

GITHUB_SSSOM:issues/166 or GITHUB_SSSOM_ISSUES:166 would be conceivable, but again, not much is gained.

My intuition of EntityReference vs URI is: If it is a URL, use a URI datatype. If it is a (semantic) identifier that refers to some domain concept, use EntityReference. Obviously, this distinction is very often fuzzy (Wikipedia page, wikidata page). I prefer it to be a URI also because this is what we have decided for ontologies in OBO!

gouttegd: > I dont think it its in the spirit of CURIEs to use them like this.

I don’t see how. The issue number is a unique identifier which only makes sense in the context of a given repository (is only unique within that repository), the URL prefix gives it a unique sense in the general context. How is that different from `FBbt:012345657`, where the local ID is only unique in the context of the Drosophila Anatomy ontology and the PURL prefix gives it global uniqueness?

> It would be: GITHUB:/issues/166, which does not really save us much..

No, it can be `GITHUB:166`, the `issues/` can be part of the prefix, why do you think it needs to be left out?

Just in case: if it’s because you want to distinguish between ”issues” and ”pull requests” (so that you can have `GITHUB:issues/166` and `GITHUB:pulls/167`), there’s no need. GitHub does _not_ distinguish between issues and pull requests, they are all ”issues”. They share the same identifier namespace. With a prefix of `https://github.com/mapping-commons/sssom/issues`, `GITHUB:166` will get you to the issue page if issue number 166 is an issue, or to the PR page if issue 166 is a PR (a PR simply being a particular case of “issue”).

> I prefer it to be a URI also because this is what we have decided for ontologies in OBO!

**Allowing** the issue ID to be curified does not **force** you to do so. If you prefer to always use full-length URL, just don’t declare a GitHub prefix. As is already the case for all other fields that have `EntityReference` as range: it has always been possible to choose _not_ to use curified values in them.

matentzn: You have not been part of a 4 year long debate about prefix harmonisation.. I will tell you the story in the next call. In essence, we are trying to promote a use of prefixes that are globally unique in a similar way as URI prefixes - even if that is practically nearly impossible.

Regardless, you are right; and I have held your position strongly in the past. But there are many people among us that would like to promote CURIEs at least in the biomedical context to be globally unique.

gouttegd: > In essence, we are trying to promote a use of prefixes that are globally unique in a similar way as URI prefixes

You mean that, for example, `FBbt` should only ever be used to refer to `http://purl.obolibrary.org/obo/FBbt_`?

I don’t doubt for a minute that there are people out there who flatly assume that is the case, and who will, for example, hardcode the decoding of such prefixes without even looking at a prefix map.

The same way that, 20 years ago, they were people who flatly assumed that the `xsl` prefix in a XML document could only refer to `http://www.w3.org/1999/XSL/Transform` without ever bothering to check the namespace declarations.

That doesn’t mean we should encourage that. I think it is a terrible idea and a recipe for disaster.

matentzn: I think this case here is borderline - issue tracker items at least on github can be considered reasonably "identifier-like". Since you are the only one engaging in the PR on the datatype front and its been open for a long time, I will go with you suggestion now.

src/sssom_schema/schema/sssom_schema.yaml
```diff
@@ -436,10 +436,10 @@ slots:
       - value: https://www.wikidata.org/wiki/Q865360
         description: (the Wikidata identifier for the Jaccard index measure).
   issue_tracker_item:
-    description: A URL location of the issue tracker item discussing this mapping or mapping set.
-    range: uri
+    description: The issue tracker item discussing this mapping.
+    range: EntityReference
     examples:
-      - value: https://github.com/mapping-commons/sssom/issues/166
+      - value: SSSOM_GITHUB_ISSUE:166
```
gouttegd, line 442: If you want to **allow** curification of the `issue_tracker_item` value without **promoting** that, you could leave the example to the full URL.

`EntityReference` is defined using LinkML’s `uriorcurie` type, so a full URL is always valid.

gouttegd (approved):

src/sssom_schema/schema/sssom_schema.yaml
matentzn: I think this is ok for now; the uriorcurie base type is another black holes we are avoiding to think about right now, so lets move this issue to a future effort.

matentzn:
I will leave this open until Monday night (31st July), then I will merge so I can add this to the OAEI report. Thanks @gouttegd for your suggestions and review!

PR #263, opened 2023-03-07
Experimental: defining "extension" slot as a list of key-value pairs

matentzn:
Resolves [#ISSUE, #ISSUE]

### DIFF
--- a/examples/embedded/uberon-external.sssom.tsv
+++ b/examples/embedded/uberon-external.sssom.tsv
@@ -1,0 +1,20 @@
+# curie_map:
+#   FBbt: http://purl.obolibrary.org/obo/FBbt_
+#   UBERON: http://purl.obolibrary.org/obo/UBERON_
+#   ZFA: http://purl.obolibrary.org/obo/ZFA_
+#   MYINTERNALNAMESPACE: http://my.internal.ns/vocab/
+#   semapv: https://w3id.org/semapv/
+#   skos: http://www.w3.org/2004/02/skos/core#
+#   dc: http://purl.org/dc/terms/
+# license: https://w3id.org/sssom/license/unspecified
+# mapping_set_id: https://w3id.org/sssom/mappings/12345287613876278135
+# extension_definition:
+#   - key: modified
+#     value: dc:modified
+#   - key: mapping_id
+#     value: MYINTERNALNAMESPACE:mapping_id
+#   - key: funded_by
+#     value: foaf:fundedBy
+subject_id	predicate_id	object_id	mapping_justification	modified	mapping_id	funded_by
+UBERON:0000004	skos:exactMatch	ZFA:0000047	semapv:UnspecifiedMatching	2022-03-01	M6972619762	NIH:RM1-HG010860-01
+UBERON:0000005	skos:exactMatch	FBbt:00005157	semapv:UnspecifiedMatching	2022-02-01	M98879876121	NIH:RM1-HG010860-01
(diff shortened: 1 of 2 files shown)


src/sssom_schema/schema/sssom_schema.yaml
```diff
@@ -459,12 +459,23 @@ slots:
   other:
     description: Pipe separated list of key value pairs for properties not part of
       the SSSOM spec. Can be used to encode additional provenance data.
-    range: string
+    range: key value pair
+    multivalued: true
+    inlined_as_list: true
```
cmungall, line 464: if this is false it makes the serialized dict slightly more idiomatic

examples/embedded/uberon-external.sssom.tsv
```diff
@@ -0,0 +1,20 @@
...
+# license: https://w3id.org/sssom/license/unspecified
+# mapping_set_id: https://w3id.org/sssom/mappings/12345287613876278135
+# extension_definition:
+#   - key: modified
+#     value: dc:modified
+#   - key: mapping_id
+#     value: MYINTERNALNAMESPACE:mapping_id
+#   - key: funded_by
+#     value: foaf:fundedBy
+subject_id	predicate_id	object_id	mapping_justification	modified	mapping_id	funded_by
```
gouttegd, line 18: What about the risk of clashes if a future version of SSSOM defines, say, `mapping_id` as a standard field?

I’d suggest mandating that columns defined by this extension mechanism should be prefixed, e.g. with `ext_` or similar.

That is, if an extension is defined as follows:

```#extension_definition:
#  - key: mapping_id
#    value: MYINTERNALNAMESPACE:mapping_id
```

Then in the TSV file it should appear as:

```
subject_id   predicate_id   [...] ext_mapping_id
```

gouttegd: Either that or adding a SSSOM version number in the mapping set metadata.

matentzn: This is exactly what I was musing about before I stopped working on it. The downside of using "ext" prefixes for column names is that I already know projects using SSSOM that really don't want to change their column names which are part of a large metadata model for mappings (with a few organisation specific columns like, ehem, species, Grant_id, and other Shananigans. I would have much liked the prefix solution for cleanliness. 

My thought was that if a custom column is defined in the mapping set header, it trumps any definitions in sssom schema. This is super messy, but it's unclear to me how I should even approach this question.

gouttegd: It may not be the cleanest solution, but it’s a pretty common way of dealing with existing legacy stuff in specifications and standards.

matentzn: This is a very French way to go about it, while the real world clearly works under Mediterranean principles.. People will just not care about this constraint in the spec - the only consequence it will have is that we can't use normal processing tools on many mapping sets in the wild. I know I sound negative but man, after 10 years of lobbying with great charme and all we barely managed to get people to agree to publish ontologies in a common syntax - it has proven nearly impossible to tell these same people to use a common set of annotation properties to describe their metadata. 

On the other hand, I cant see a different solution either! I am 51% inclined to go your way. I wouldn't even bother to enumerate existing columns, they are too messy anyways. Just force the `ext_`.

gouttegd: This _is_ a real-world approach. Look at any IETF RFCs, any IANA registry: they are littered with this kind of “reserved” attributes, keywords, identifiers. Most of them exist for only one reason: to deal with this kind of mess and the fact that many people don’t give a shit about standards, and standards have to take that into account.

matentzn:
Superceded by #375

PR #266, opened 2023-03-11
Make adding a concrete SSSOM example part of the new element request

matentzn:
Resolves [#ISSUE, #ISSUE]


cthoyt (approved):
This is much appreciated - based on the feedback I gave in https://github.com/mapping-commons/sssom/pull/258, this will be helpful.

However, before we finish this one, can we consider how we might operationalize this? Like can we have in each field's definition a link to a list of valid SSSOM files that include the example element?


Note: I made a small cosmetic change to the table so it displays properly.

matentzn:
@cthoyt I added a commend requiring the linking of issues related to a metaelement into the linkml model itself. Sounds good?

cthoyt:
@matentzn amazing! Can we retroactively add an example that we can point to so people know how to enter something in this field correctly?

PR #267, opened 2023-03-14
Add some examples of using see_also to SSSOM schema

matentzn:
See also #266 

This PR adds a bunch of examples of using the `see_also` property correctly. I also checked if it works - they show up in the RDF serialisations of the LinkML schema.

### DIFF
--- a/src/sssom_schema/schema/sssom_schema.yaml
+++ b/src/sssom_schema/schema/sssom_schema.yaml
@@ -20,6 +20,7 @@ default_curi_maps:
 - semweb_context
 - obo_context
 default_prefix: sssom
+default_range: string
 
 enums:
   entity_type_enum:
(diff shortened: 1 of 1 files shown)


src/sssom_schema/schema/sssom_schema.yaml
```diff
@@ -134,6 +135,9 @@ slots:
       optional field is documentation for human reviewers - when a category is known
       and documented clearly, the cost of interpreting and evaluating the mapping decreases.
     range: string
+    see_also:
+      - https://github.com/mapping-commons/sssom/issues/13
```
cthoyt, line 139: While linking to related issues talking about fields as they were added, this is not what I was talking about. I want to see a link to a TSV file that demonstrates actual usage of the field in an actual SSSOM file.

cthoyt:
Also please update the text in the issue template to more strongly suggest giving an example as a TSV

src/sssom_schema/schema/sssom_schema.yaml
matentzn: I added an example for this to the #258 pull request, with a link to an SSSOM file. We will make adding a properly formatted SSSOM file a requirement, but I don't think it has be be real. The discussion in the issue will have to make a case it is needed - the example should merely illustrate how it is intended to be used.

matentzn:
> Also please update the text in the issue template to more strongly suggest giving an example as a TSV

We will make it required.

PR #268, opened 2023-03-14
Adding QC checks for example SSSOM files

matentzn:
This pull request adds continuous integration testing for all example SSSOM files provided in the `examples` directory, see also @cthoyt comments in #258.


.github/workflows/qc.yml
```diff
@@ -0,0 +1,31 @@
...
+# A workflow run is made up of one or more jobs that can run sequentially or in parallel
+jobs:
+  sssom_validation:
+    runs-on: ubuntu-latest
+    container: obolibrary/odkfull:v1.4
+    steps:
+      - name: Install latest SSSOM
+        env:
+          DEFAULT_BRANCH: master
+        run: pip install --upgrade pip && pip install -U sssom
```
cthoyt, line 25: should this not install SSSOM locally from the repository that's been pulled, so it can be used to test changes to the code as well? Otherwise, this PR is great.

matentzn: This requires more work, because the sssom toolkit is in a different location, and maintained differently then the stuff maintained here. We know what to do: https://github.com/mapping-commons/sssom/issues/115

But we have not gotten around to this yet.

PR #269, opened 2023-03-14
Making adding a link to a working example required when adding a new field

matentzn:
This PR extends the pull request template to make adding a working example to a metadata field required.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -9,6 +9,7 @@ If you are proposing a change to the SSSOM metadata model, you must
 
 - [ ] provide a full, working and valid example in `examples/`
 - [ ] provide a link to the related GitHub issue in the `see_also` field of the linkml model
+- [ ] provide a link to a the valid example in the `see_also` field of the linkml model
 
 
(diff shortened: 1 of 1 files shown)


matentzn:
@cthoyt lets not get too hung up now on the "realness" of the example. The request must be believable, but if it is real or not is a question that goes a bit far as a requirement IMO. If the request is not credible enough, then we wont even make a PR for it (its not that anyone else ever would).

.github/pull_request_template.md
```diff
@@ -9,6 +9,7 @@ If you are proposing a change to the SSSOM metadata model, you must
 
 - [ ] provide a full, working and valid example in `examples/`
 - [ ] provide a link to the related GitHub issue in the `see_also` field of the linkml model
+- [ ] provide a link to a the valid example in the `see_also` field of the linkml model
```
cthoyt, line 12:
```suggestion
- [ ] provide a link to a valid example in the `see_also` field of the linkml model
```

cthoyt (approved):

PR #277, opened 2023-04-27
Cookiecutter directions for new project generation

### LINKED ISSUE #273: Document how to create a new mapping commons using cookiecutter
@hrshdhgd would you be able to add a documentation page here https://github.com/mapping-commons/sssom/tree/master/src/docs called `create-mapping-commons.md` and fill it with some initial instructions and 

1. how to create a new mapping commons from zero to hero and
2. how to update one

We have a new team member (@ehartley) who will help us building the template up.

hrshdhgd:
Resolves [#ISSUE, #ISSUE]

Fixes #273


matentzn (approved):
Ok that looks good to get this started!

src/docs/create-mapping-commons.md
```diff
@@ -0,0 +1,47 @@
+# Getting started
+
+- First, create a virtual environment of your choice (anaconda, venv, poetry etc.).
```
ehartley, line 3: I think less technical users will need some guidance on how to create a virtual environment. However, with the numerous options available, I'm not sure whether it would be better to provide an example or 2 or to just include some reference links.

hrshdhgd: Good call! I've updated the documentation with a 'best practices guide' that our group uses.

ehartley:
One last suggestion, I think these instructions should also include a section on how to publish the project to GitHub/GitLab.

matentzn:
@hrshdhgd can you finalise this PR?

hrshdhgd:
I've added the instructions for GitHub. I've never used gitlab so someone else could fill in the blanks there. But from my perspective this should be ready to merge if the `TODO`s are acceptable to fill in later.

matentzn (approved):
Thanks @hrshdhgd I will call this good enough for now, @ehartley if you want clarifications for anything keep make issues, we will deal with them one by one!

Issue #283, opened 2023-06-10
Issues with the specification “overview”


gouttegd:
The [Overview of the SSSOM specification](https://mapping-commons.github.io/sssom/spec/) suffers from some issues.

1. It allows to overload the value of the `curie_map` parameter (in the YAML metadata block, when serialising in TSV) so that it can be “either dictionary of CURIE->URLPREFIX pairs or a link to a valid curie map of the same shape”. There is agreement on Slack that such overloading is not a good idea.

The document should be amended so that `curie_map` can only be a dictionary of {CURIE, URLPREFIX} pairs. If the ability to reference an external curie map is still desired, a new parameter should be introduced (for example, `curie_map_reference`, `curie_map_link`, or something like that).

(For what it’s worth, I am slightly against allowing a reference to an external curie map, but I don’t have a strong opinion on that issue. I do think allowing such references would make things more complicated that they need to be, but I can deal with that.)

2. The document is inconsistent regarding the filename convention for the external metadata file. It first recommends `.sssom.yml`

> External yaml metadata files should have the extension `.sssom.yml`, for example `mp-hp-exact-0.0.1.sssom.tsv`

(of note, there’s already an inconsistency here, as the given example has a `.sssom.tsv` extension), then it recommends `-meta.yml`

> In external mode, the mapping set metadata is supplied by a separate YAML file having the same base-name of the mapping file, with the extension `-meta.yml`.

The real recommended filename should be `.sssom.yml`.

3. The sample TSV file violates the stated requirement on column orders. The description of the TSV format says:

> The columns MUST be sorted according to the order as they appear in the SSSOM metadata. For example, the first columns of a mapping set TSV should always be, in that order: subject_id, predicate_id, object_id, mapping_justification, if labels are not included; if they are included, the order should be: subject_id, subject_label, predicate_id, predicate_label, object_id, object_label, mapping_justification.

The [example file](https://raw.githubusercontent.com/mapping-commons/sssom/master/examples/external/mp-hp-exact-0.0.1.sssom.tsv) uses the following ordering: subject_id, predicate_id, object_id, mapping_justification, subject_label, object_label.

This begs the question: is column ordering really important? If it is, then we should not provide an example that does not follow it. If it is not (e.g., if it’s only a recommendation, as row ordering seems to be), then the MUST should be relaxed at least into a SHOULD.

And speaking of MUST and SHOULD:

[...]

PR #285, opened 2023-06-11
Issues around canonical ordering

matentzn:
Partially adresses #283 

- Applies canonical sort ordering to all examples files
- Downgrades the necessity of applying the canonical column order from MUST to SHOULD
- Documents clearly that built-in prefixes MUST NOT be redefined


examples/external/mp-hp-exact-0.0.1.sssom.tsv
matentzn, line 1: Looks like a big change but I just use the same data as in the embedded version.

examples/embedded/foodie-inc-2022-05-01.sssom.tsv
```diff
@@ -1,25 +1,30 @@
+# comment: We could map to FOODON:00004187 instead which more specifically refers to
+#   'raw' Pink apples. Decided against to be consistent with other mapping choices.
 # curie_map:
 #   FOODON: http://purl.obolibrary.org/obo/FOODON_
 #   KF_FOOD: https://kewl-foodie.inc/food/
+#   orcid: https://orcid.org/
+#   owl: http://www.w3.org/2002/07/owl#
+#   rdf: http://www.w3.org/1999/02/22-rdf-syntax-ns#
+#   rdfs: http://www.w3.org/2000/01/rdf-schema#
+#   semapv: https://w3id.org/semapv/
```
gouttegd, line 10: Wait… Is the URL for `semapv` `https://w3id.org/semapv/` or `https://w3id.org/semapv/vocab/`???

All SEMAPV IDs in [the SEMAPV repository](https://github.com/mapping-commons/semantic-mapping-vocabulary/blob/main/semapv.owl) are of the form `https://w3id.org/semapv/vocab/XXX` (e.g. `https://w3id.org/semapv/vocab/LexicalMatching`).

src/docs/spec.md
```diff
@@ -151,7 +151,7 @@ Note this is a public copy of the editors’ draft. It is provided for discussio
 
 ## SSSOM Metadata Elements
 
-The SSSOM specification defines a set of SSSOM metadata elements that are used to describe mappings. Apart from the elements themselves, some example usage and a description, **_[the SSSOM spec](https://mapping-commons.github.io/sssom/Mapping/) defines the canonical order for the metadata_** in which the elements should appear when serialised. This precludes spurious diffs in a git setting, which is an important concern for the continuous reviewing of mappings by curators and users. 
+The SSSOM specification defines a set of SSSOM metadata elements that are used to describe mappings. Apart from the elements themselves, some example usage and a description, **_[the SSSOM spec](https://mapping-commons.github.io/sssom/Mapping/) defines the canonical order for the metadata_** in which the elements SHOULD appear when serialised. This precludes spurious diffs in a git setting, which is an important concern for the continuous reviewing of mappings by curators and users.
```
gouttegd, line 154: I would actually move that entire paragraph. It only concerns serialisation, so it should be in the section that describe how SSSOM mapping sets are serialized.

src/docs/spec.md
```diff
@@ -351,9 +351,10 @@ SubClassOf(Annotation(sssom:creator_id <https://orcid.org/0000-0002-7356-1779>)
...
 
-Metadata about a set of mappings can be supplied as part of the mappings (embedded mode) and as a simple yaml file alongside the primary mapping file. Note that for the TSV, it will be required to supply a valid curie map that allows the unambiguous interpretation of CURIEs. A curie map is supplied after a `curie_map:` parameter in the yaml file. The value is a dictionary of CURIE->URLPREFIX pairs.
+Metadata about a set of mappings can be supplied as part of the mappings (embedded mode) and as a simple yaml file alongside the primary mapping file. The YAML metadata block MUST contain a curie map that allows the unambiguous interpretation of CURIES. A curie map is supplied after a `curie_map:` parameter in the yaml file. The value is a dictionary of CURIE->URLPREFIX pairs.
+Note that the following prefixes are built-in and (1) MUST NOT be changed from their bioregistry conformant default interpretation and (2) MAY be omitted from the curie map: "`sssom`", "`owl`", "`rdf`", "`rdfs`", "`skos`", "`semapv`".
```
gouttegd, line 357: It would be better to provide directly a table with the URL corresponding to those prefixes, rather than delegating to the bioregistry. At the very least, if we do delegate to the bioregistry we should provide a reference and/or a link (don’t assume implementers will necessarily know that ”bioregistry” refers to).

gouttegd (approved):
Approve, but please note the remark about the SEMAPV prefix URL.  The one given in the examples’ curie map differs from the one used in SEMAPV and the one registered on the bioregistry.

examples/embedded/foodie-inc-2022-05-01.sssom.tsv
matentzn: That is a big bug! Thanks for noticing. https://github.com/mapping-commons/sssom/pull/287

The model level prefix map overwrote the bioregistry context!

src/docs/spec.md
matentzn: Agreed and done.

src/docs/spec.md
matentzn:
```suggestion
Note that the following prefixes are built-in and (1) MUST NOT be changed from their [SSSOM default interpretation](https://github.com/mapping-commons/sssom/blob/master/project/jsonld/sssom_schema.context.jsonld) and (2) MAY be omitted from the curie map: "`sssom`", "`owl`", "`rdf`", "`rdfs`", "`skos`", "`semapv`".
```

matentzn: And it was also wrong!

examples/embedded/foodie-inc-2022-05-01.sssom.tsv
matentzn: Also fixed now in this PR

gouttegd (approved):

matentzn:
- [ ] Requires a new sssom-py release @hrshdhgd (needed a sssom schema updated, which is needed by the sssom.context refresh which depends on sssom_Schema release... BRAIN TEASE). Can you keep an eye on this?

hrshdhgd:
@matentzn , I noticed that the `qc.yml` workflow here is using `odkfull:v1.4` docker image. I think if you update the `sssom` version in the image to the latest version, this should pass. Does this make sense?

hrshdhgd:
> Does it make sense, instead to add a pip install -U sssom to the action to ensure its _always_ on the most up to date state?

I agree! That's the ideal scenario.

hrshdhgd:
But does the Docker image already have the latest version of `sssom`. I made a release just yesterday.

matentzn:
@hrshdhgd feel free to merge, its fine!

PR #287, opened 2023-06-11
Fix semapv namespace in SSSOM schema

matentzn:
The semapv namespace is wrongly documented in the LinkML model.

Instead of `https://w3id.org/semapv/` it should be `https://w3id.org/semapv/vocab`, see SEMAPV:

https://github.com/mapping-commons/semantic-mapping-vocabulary/blob/main/semapv.owl

e.g.:

https://w3id.org/semapv/vocab/CaseNormalization


matentzn:
@gouttegd this happens when you manage prefixes in multiple places.. :P

PR #289, opened 2023-06-14
Package context files through `make`

hrshdhgd:
This enables packages that use `sssom-schema` as a dependency to use context files within this package rather than doing a `wget` which is not efficient.

This will be address : https://github.com/mapping-commons/sssom-py/issues/375 via https://github.com/mapping-commons/sssom-py/pull/382

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -40,7 +40,7 @@ install:
 	poetry install
 .PHONY: install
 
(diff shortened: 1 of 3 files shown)


matentzn (approved):
I am fine with this! Thanks!

PR #294, opened 2023-07-04
Fix publication_date

### LINKED ISSUE #293: Missing property on web page
Go to : 
https://mapping-commons.github.io/sssom/

publication_date is referred here: 
https://mapping-commons.github.io/sssom/publication_date/

However, the properties is not used by any of the objects described on the page (Mapping, MapppingSet, MappingRegistry, MappingSetReference).

matentzn:
Resolves #293 

A property (publication_date) was defined in SSSOM, but was not added to the SSSOM mapping and mapping set data model. Since a very similar property (mapping_date) was already defined, this change is considered a bugfix, rather than an added feature.

### DIFF
--- a/src/sssom_schema/schema/sssom_schema.yaml
+++ b/src/sssom_schema/schema/sssom_schema.yaml
@@ -522,6 +522,7 @@ classes:
     - mapping_provider
     - mapping_tool
     - mapping_date
+    - publication_date
     - subject_match_field
     - object_match_field
     - subject_preprocessing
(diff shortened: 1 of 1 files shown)


gouttegd (approved):
To be honest I am not convinced it makes sense to have a `publication_date` at the mapping level, anymore than I am convinced it makes sense to have a `mapping_date` at the mapping set level.

To me it would make more sense to have only a `publication_date` at the mapping set level (“when the mapping set was published”) and a `mapping_date` at the mapping level (“when that particular mapping was asserted”).

But you know SSSOM use cases better than I do, so I assume you know what you’re doing. ;)

PR #302, opened 2023-07-25
Fix SSSOM Schema not having bijective prefix map

cthoyt:
There are issues in https://github.com/mapping-commons/sssom-py/pull/397 because the base schema does not have a valid prefix map - `dc` and `dcterms` are both used as prefixes for the same URI prefix. This happens because LinkML makes some inference about what needs to be there - there are some assorted uses throughout the schema definition with both DC and DCTERMS. This PR updates the schema to only use DCTERMS and puts explicit entries for DCTERMS in the prefix map to address this. It then regenerates the whole project.


matentzn (approved):

matentzn:
@hrshdhgd When you tink this is good to go, can you merge it and make a new release?

hrshdhgd:
Thanks a lot @cthoyt !!

@matentzn , I ran `make all`.

cthoyt:
everything in this is looking good now.

The one (non-blocking) exception is the shex output, which appears to have a bug in the way that it propagates the context (I think it assumes DC is used implicitly). This is not a blocker, since this is still a valid, self-contained document. The only issue is aesthetics.

matentzn (approved):
lgtm

PR #303, opened 2023-07-25
Run `make all`

cthoyt:
This PR does not change the underlying schema, it just runs `make all`

This also motivates some kind of CI check that when the schema is updated, then this gets run.

I want to do this to reduce spurious diff in #302


matentzn (approved):
No issues here. @hrshdhgd you can merge

PR #306, opened 2023-07-27
Clarifying descriptions of various properties

### LINKED ISSUE #304: Do not allow free-form text in EntityReference fields
The description of the `creator_id` slot states:

> Identifies the persons or groups responsible for the creation of the mapping. […] Recommended to be a (pipe-separated) list of ORCIDs or otherwise identifying URLs, but any identifying string (such as name and affiliation) is permissible.

This seems to allow the use of free-form non-identifiers (e.g. “A.U. Thor, Institute of Imprecise Mappings, Shangri-La”). I believe this should _not_ be allowed. If all that is available to “identify” the creator of a mapping is such a string, it should go to the `creator_label` slot instead.

`creator_id` is supposed to be an `EntityReference`, which itself should be a `uriorcurie`. The documentation should not encourage mapping creators to fill that field with a value that does not match that range.

(As an aside: the “(pipe-separated)“ also has nothing do to here, in my opinion. It’s a serialisation issue, which does not belong in a description of the model.)

matentzn:
Resolves #304 

In this PR, we 

- clarify the scope of a number of SSSOM properties (e.g author, reviewer and creator_id properties are clarified to be URIs, rather than "strings")
- Drop the use of syntactic concerns such "pipe-seperated" which do not belong in a general spec
- fix some minor typos

### DIFF
--- a/src/sssom_schema/schema/sssom_schema.yaml
+++ b/src/sssom_schema/schema/sssom_schema.yaml
@@ -487,7 +487,7 @@ slots:
     range: string
     multivalued: true
   other:
-    description: List of key-value pairs for properties not part of
+    description: Pipe separated list of key value pairs for properties not part of
       the SSSOM spec. Can be used to encode additional provenance data.
     range: string
   comment:
--- a/src/sssom_schema/schema/sssom_schema.yaml
+++ b/src/sssom_schema/schema/sssom_schema.yaml
@@ -298,40 +298,37 @@ slots:
       the mapping. The creator is the agent that put the mapping in its published form, 
       which may be different from the author, which is a person that was actively involved
       in the assertion of the mapping.
-      Recommended to be a (pipe-separated) list of ORCIDs or otherwise
-      identifying URLs, but any identifying string (such as name and affiliation)
-      is permissible.
+      Recommended to be a list of ORCIDs or otherwise
+      identifying URIs.
     slot_uri: dcterms:creator
     range: EntityReference
     multivalued: true
   creator_label:
     description: A string identifying the creator of this mapping. In the spirit of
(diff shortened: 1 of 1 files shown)


src/sssom_schema/schema/sssom_schema.yaml
```diff
@@ -490,7 +487,7 @@ slots:
     range: string
     multivalued: true
   other:
-    description: Pipe separated list of key value pairs for properties not part of
+    description: List of key-value pairs for properties not part of
```
gouttegd, line 490: 1. If that slot is intended to be a list, then it should be `multivalued: true`.
2. Somewhere in the spec you need to specify how those key-value pairs are represented. I suppose they should be of the form `key=value`, but you don’t want to leave implementers guessing. (Unless Linkml already has a type for key-value pairs, in which case that type should be used here instead of `string`.)

gouttegd (changes requested):

src/sssom_schema/schema/sssom_schema.yaml
gouttegd: To clarify: the change I’m requesting for this PR is the first one (setting the slot to be multivalued). I don’t mind if the details of the serialisation of key-value pairs are left to be fixed later in another PR.

gouttegd: Having key pairs formally represented in the model is indeed the best solution. I didn’t suggest it because I thought it would be too big of a change, I hadn’t noticed there was already a PR for that.

gouttegd: However, this whole “extension” thing is going to be a nightmare for implementations in strongly typed languages…

PR #317, opened 2023-09-13
Change sempav to semapv in the tutorial page

jmillanacosta:
Hi mapping commons, here with a suggestion for the tutorial page. 

Googled to see which is the recommended prefix but I'm still not sure to which one you will stick. The different spelling confused me a little bit when following the tutorial, is this something you would like to address?


matentzn (approved):
Fantastic, thanks for the drive-by @jmillanacosta!

PR #319, opened 2023-09-15
Recommend that YAML keys be alphabetically sorted.

gouttegd:
This PR adds a recommendation that the keys in the YAML metadata block (in both "embedded" and "external" modes) SHOULD be sorted by case-sensitive alphabetical order.

There is no particular reason for recommending alphabetical order over any other type of order, beyond the fact that `sssom-py` is already using alphabetical order (probably because that is seemingly the default behaviour of the PyYAML's `safe_dump` method used by `sssom-py`). So we just take the order that is _de facto_ in use in the wild and make it the officially recommended order.

- [ ] `make test` has been run locally (no, documentation change only)
- [ ] tests have been added/updated (no, documentation change only)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Next
 
 - Add issue_tracker_item and issue_tracker [model elements](https://github.com/mapping-commons/sssom/pull/259).
+- Add recommendation to sort the keys in the YAML metadata block.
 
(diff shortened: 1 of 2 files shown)


gouttegd:
Closing since the alternative proposition (#320) has been approved. (In the event someone objects to “spec order“ and makes a case for alphabetical order we can reopen this one.)

PR #320, opened 2023-09-18
Recommend that YAML keys be sorted by "spec order".

gouttegd:
This PR adds a recommendation that

1) the keys in the YAML metadata block (in both "embedded" and “external” modes) SHOULD be sorted in the same order as the corresponding slots are listed in the definition of the [MappingSet class](https://mapping-commons.github.io/sssom/MappingSet/) in the SSSOM model;

2) the keys in the `curie_map` dictionary (i.e. the prefix names) SHOULD be sorted by alphabetical order.

This is an alternative PR to #319 (which proposed to sort the YAML metadata block by alphabetical order). “Spec order“ is not the current behaviour of `sssom-py`, but is preferred at least by @matentzn.

Of note, since `curie_map` is currently _not_ part of the model (and therefore not listed in the definition of `MappingSet`), the position of the `curie_map` key would be unspecified under the proposed recommendation. For what it’s worth, SSSOM-Java puts it at the top of the metadata block, before any other slot.

- [ ] `make test` has been run locally (no, irrelevant)
- [ ] tests have been added/updated (no, irrelevant)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Next
 
 - Add issue_tracker_item and issue_tracker [model elements](https://github.com/mapping-commons/sssom/pull/259).
+- Add recommendation to sort the keys in the YAML metadata block.
 
 ## SSSOM version 0.13.0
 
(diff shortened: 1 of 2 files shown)


matentzn (approved):
Let's go with this! Thank you!

matentzn:
This can be merged on Friday 22nd by anyone who sees this without permission! I just leave it Open in case there are objections that I don't know about.

PR #347, opened 2024-01-31
Add new slot "mapping_status" (mapping lifecycle) to SSSOM core model

### LINKED ISSUE #345: [New metadata element]: status

```
status
```

```
URI
```

**Description**
Indicate the status of this individual mapping. This allows a submission/review/publication/retirement lifecycle to be tracked using a set of mappings with the same subject/predicate/object. 

Standard status values should be established for SSSOM. I suggest
- draft [1..*]
- submitted 
- reviewed [1..*]
- published
- withdrawn

Example description.

```
# curie_map:
#   owl: http://www.w3.org/2002/07/owl#
#   rdf: http://www.w3.org/1999/02/22-rdf-syntax-ns#
#   rdfs: http://www.w3.org/2000/01/rdf-schema#
#   sssom: https://w3id.org/sssom/
#   orcid: https://orcid.org/
#   status: https://w3id.org/sssom/status/
#   get: https://global-ecosystems.org/explore/
#   alum8: http://www.neii.gov.au/def/voc/ACLUMP/australian-land-use-and-management-classification/
# license: https://creativecommons.org/publicdomain/zero/1.0/
# mapping_set_id: http://w3id.org/env/crosswalk/alum2get/2024-01-13
```

| subject_id	| predicate_id	  | object_id	  | mapping_justification   | author_id	|  mapping_date | status | comment |
| ----------- | --------------- | ----------- | ----------------------- | ------------------- | -- | -- | -- |
| alum8:Marsh-or-wetlandsaline	| skos:exactMatch	| get:groups/MFT1.3	| semapv:ManualMappingCuration	| orcid:0009-0001-6090-9959	|  2023-12-01  |  status:submitted |  |
| alum8:Marsh-or-wetlandsaline	| skos:exactMatch	| get:groups/MFT1.3	| 	| orcid:0000-0002-2934-5497	|  2024-01-03  |  status:reviewed | Looks Ok but let's check with Piers |
| alum8:Marsh-or-wetlandsaline	| skos:exactMatch	| get:groups/MFT1.3	| 	| orcid:0000-0002-2568-59457	|  2024-01-10  |  status:reviewed | approved | 
| alum8:Marsh-or-wetlandsaline	| skos:exactMatch	| get:groups/MFT1.3	| 	| orcid:0000-0002-2934-5497	|  2024-01-13  |  status:published |  |

matentzn:
Resolves #345 

This is a draft PR adding a `mapping_status` metadata field to the SSSOM core model to describe the life cycle phase a specific mapping is in. Please use the issue for general discussions, and the PR only about specific suggestions.

### DIFF
--- a/examples/schema/mapping_status.sssom.tsv
+++ b/examples/schema/mapping_status.sssom.tsv
@@ -1,0 +1,20 @@
+#curie_map:
+#  owl: http://www.w3.org/2002/07/owl#
+#  rdf: http://www.w3.org/1999/02/22-rdf-syntax-ns#
+#  rdfs: http://www.w3.org/2000/01/rdf-schema#
+#  semapv: https://w3id.org/semapv/vocab/
+#  skos: http://www.w3.org/2004/02/skos/core#
+#  sssom: https://w3id.org/sssom/
+#  orcid: https://orcid.org/
+#  get: https://global-ecosystems.org/explore/
+#  alum8: http://www.neii.gov.au/def/voc/ACLUMP/australian-land-use-and-management-classification/
+#mapping_set_id: https://w3id.org/sssom/commons/examples/mapping_status.sssom.tsv
+#license: "https://creativecommons.org/publicdomain/zero/1.0/"
+#creator_id: orcid:0000-0002-3884-3420
+#mapping_provider: "https://w3id.org/sssom/core_team"
+#comment: This is an example file for the SSSOM for illustration only. Its contents are entirely fabricated.
+subject_id	predicate_id	object_id	mapping_justification	author_id	mapping_date	status	comment
+alum8:Marsh-or-wetlandsaline	skos:exactMatch	get:groups/MFT1.3	semapv:ManualMappingCuration	orcid:0009-0001-6090-9959	2023-12-01	draft	
+alum8:Marsh-or-wetlandsaline	skos:exactMatch	get:groups/MFT1.3	semapv:ManualMappingCuration	orcid:0000-0002-2934-5497	2024-01-03	reviewed	Looks Ok but let's check with Piers
+alum8:Marsh-or-wetlandsaline	skos:exactMatch	get:groups/MFT1.3	semapv:ManualMappingCuration	orcid:0000-0002-2568-59457	2024-01-10	approved	Mapping has been approved by Pier and team.
+alum8:Marsh-or-wetlandsaline	skos:exactMatch	get:groups/MFT1.3	semapv:ManualMappingCuration	orcid:0000-0002-2934-5497	2024-01-13	published	
--- a/src/sssom_schema/schema/sssom_schema.yaml
+++ b/src/sssom_schema/schema/sssom_schema.yaml
@@ -60,6 +60,21 @@ enums:
       "1:0": One-to-none mapping
       "0:1": None-to-one mapping
       "n:n": Many-to-many mapping
+  status_enum:
+    permissible_values:
+      draft:
+        description: "Indicates the mapping is in an early, unrefined form. This status is typically used for works in progress that are not yet ready for formal review or publication."
+      submitted:
+        description: "Represents the stage where the mapping has been submitted for review but has not yet been evaluated. This status is used after initial development and indicates readiness for external assessment."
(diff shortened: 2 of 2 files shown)


examples/schema/mapping_status.sssom.tsv
```diff
@@ -0,0 +1,20 @@
...
+#  alum8: http://www.neii.gov.au/def/voc/ACLUMP/australian-land-use-and-management-classification/
+#mapping_set_id: https://w3id.org/sssom/commons/examples/mapping_status.sssom.tsv
+#license: "https://creativecommons.org/publicdomain/zero/1.0/"
+#creator_id: orcid:0000-0002-3884-3420
+#mapping_provider: "https://w3id.org/sssom/core_team"
+#comment: This is an example file for the SSSOM for illustration only. Its contents are entirely fabricated.
+subject_id	predicate_id	object_id	mapping_justification	author_id	mapping_date	status	comment
+alum8:Marsh-or-wetlandsaline	skos:exactMatch	get:groups/MFT1.3	semapv:ManualMappingCuration	orcid:0009-0001-6090-9959	2023-12-01	draft	
+alum8:Marsh-or-wetlandsaline	skos:exactMatch	get:groups/MFT1.3	semapv:ManualMappingCuration	orcid:0000-0002-2934-5497	2024-01-03	reviewed	Looks Ok but let's check with Piers
+alum8:Marsh-or-wetlandsaline	skos:exactMatch	get:groups/MFT1.3	semapv:ManualMappingCuration	orcid:0000-0002-2568-59457	2024-01-10	approved	Mapping has been approved by Pier and team.
```
cthoyt, line 19: I don't get what's the difference between approved and published based on this example, also this doesn't match the enum values

examples/schema/mapping_status.sssom.tsv
```diff
@@ -0,0 +1,20 @@
...
+alum8:Marsh-or-wetlandsaline	skos:exactMatch	get:groups/MFT1.3	semapv:ManualMappingCuration	orcid:0000-0002-2934-5497	2024-01-13	published
```
cthoyt, line 20: if the mapping is in a document, then it's published, so not really sure why this even exists

cthoyt:
Overall I think this is a weak proposal. It increases complexity of the format while being sort of redundant of several things that already exist. Why not use some of the fields where you can put whatever arbitrary metadata you want to facilitate this for the minority of users who might want to use it?

I think we will keep getting more and more requests to add arbitrary enums/metadata fields based on single use cases, and I want to oppose these

gouttegd:
My main problem with this is that it’s not merely just an addition of a field (that in itself wouldn’t be a big deal), it’s an addition that changes a rather important aspect of the format.

Until now, any single mapping record (i.e. “row” in the TSV format) is “standalone” – the record contains all the metadata needed to evaluate it and decide what to do with it. No need to look anywhere else in the mapping set to make a decision.

With that change, for every mapping I now need to check if there’s _another_ record further down in the set for the same triple {subject, predicate, object} but with a different status (in particular, with a status of `withdrawn`, which would invalidate a previous record with a status of `published`).

I am not fundamentally opposed to that, but I do feel this requires more consideration. In particular, if the idea is to “provide the metadata for the client to make the decision about "which records are valid"”, then the specification needs to be much more clear about how that decision of “which records are valid” should be made. Otherwise we run the risk of ending up with a format that is interoperable in name only, because each implementation will have its own logic to determine the validity of records.

examples/schema/mapping_status.sssom.tsv
dr-shorthair: There is commonly a delay between approval and the actual release, which may be phased with other items or on pre-defined calendar dates.

examples/schema/mapping_status.sssom.tsv
dr-shorthair: That view may hold in amongst developers, but not in the statutory (government) sector or even amongst academics (who regularly withhold information 'until the journal article appears').

matentzn:
A bit unfortunate, but closing this PR here. Please move discussion back to https://github.com/mapping-commons/sssom/issues/345

PR #353, opened 2024-03-28
upgrade `mkdocs-mermaid2-plugin` major version

sujaypatil96:
This PR upgrades the major version of the `mkdocs-mermaid2-plugin` package which is asserted as a dependency on this library. The reason why this is necessary is because [oaklib](https://github.com/INCATools/ontology-access-kit/) depends on `sssom-schema` and other downstream packages which use `oaklib` as a dependency are not able to enforce the latest versions of `mkdocs-mermaid2-plugin` in their repos because of conflicting dependency versions.


hrshdhgd (dismissed):

poetry.lock
```diff
@@ -2043,77 +1948,57 @@ dev = ["flake8", "flake8-annotations", "flake8-bandit", "flake8-bugbear", "flake
...
-platformdirs = ">=3.9.1,<4"
-
-[package.extras]
-docs = ["furo (>=2023.7.26)", "proselint (>=0.13)", "sphinx (>=7.1.2)", "sphinx-argparse (>=0.4)", "sphinxcontrib-towncrier (>=0.2.1a0)", "towncrier (>=23.6)"]
-test = ["covdefaults (>=2.3)", "coverage (>=7.2.7)", "coverage-enable-subprocess (>=1)", "flaky (>=3.7)", "packaging (>=23.1)", "pytest (>=7.4)", "pytest-env (>=0.8.2)", "pytest-freezer (>=0.4.8)", "pytest-mock (>=3.11.1)", "pytest-randomly (>=3.12)", "pytest-timeout (>=2.1)", "setuptools (>=68)", "time-machine (>=2.10)"]
-
 [[package]]
 name = "watchdog"
-version = "3.0.0"
+version = "4.0.0"
```
hrshdhgd, line 1968: This will cause downstream errors specifically in GitHub actions. Keep this at 3.0.0

hrshdhgd (approved):

poetry.lock
sujaypatil96: Thanks @hrshdhgd! i've updated the `poetry.lock` such that it is updated only based on running:

```bash
poetry add mkdocs-mermaid2-plugin@latest
```

Issue #330, opened 2023-11-09
The spec/doc should be reorganised


gouttegd:
Right now, the SSSOM website is a bit of a mess.

* The [home page](https://mapping-commons.github.io/sssom/home/) and the [about page](https://mapping-commons.github.io/sssom/about/) are mostly redundant.
* The [specification page](https://mapping-commons.github.io/sssom/) is merely the auto-generated documentation of the underlying LinkML model – calling that a “specification“ is misleading as this is woefully insufficient to anyone willing to implement the standard (necessary, sure, but not sufficient).
* The [overview page](https://mapping-commons.github.io/sssom/spec/) (which, also misleadingly, is under the URL `/sssom/spec/`) is a bit of everything:
  * a list of contributors (nothing wrong with that but maybe could get its own page, or be added to the existing [credits page](https://mapping-commons.github.io/sssom/credits/);
  * a general introduction to the concept of mapping;
  * a fleeting reference to the data model;
  * a list of the commonly used and recommended mapping predicates;
  * an attempt at formally specifying the OWL/RDF and SSSOM/TSV serialisation formats;
  * a list of use cases.
* The “resources for users“ section it itself a mix of many things. Strikingly, it contains bit that really belong to the “specification” part, such as this (in the [basic tutorial](https://mapping-commons.github.io/sssom/tutorial/#curie)):

> All three must be referred to by an identifier in CURIE syntax ([Compact URI](https://www.w3.org/TR/2010/NOTE-curie-20101216/)) when using the SSSOM table format or JSON, or an IRI (Internationalized Resource Identifier) when you are using the RDF representation of SSSOM.

Overall this makes it very difficult for implementers to figure out what and where are the really “normative“ parts of the website. This is of course at least partially due to the fact that the website is clearly a “work in progress“, but also, more fundamentally, because the website somehow tries to be simultaneously a _specification_ for the standard, a _documentation_ of that standard for end-users, a documentation of the reference implementation (`sssom-py`), and an _academic paper_ on semantic mappings.

More immediately, this makes it difficult for me to figure out where I should put the various improvements to the spec I have been thinking about (such as the propagation of metadata slots #305, the recommendations on backwards compatibility #325, or the recommendations on how to deal with non-standard slots #328).

Therefore I’d like to propose that the website be reorganised so as to clearly separate the _specification_, the _documentation_, and the _general notions on mappings_. I welcome any suggestion for a better organisation, but right now here’s what I’m considering:

[...]

PR #358, opened 2024-04-08
Update documentation for entity reference to clarify CURIE/URI type

matentzn:
Resolves https://github.com/mapping-commons/sssom-py/issues/512

The documentation for "entity references" is largely confusing, and leading, in its current form, to the expectation that metadata fields types with these can be _either_ a CURIE _or_ a full URI.

This is not quite right: in the TSV serialistion, as well as JSON, we expect the values to be CURIE-strings. This PR documents this design decision, but, as @joeflack4 puts it rightfully, this has a bit of a "smell" to it. Not sure what the right thing is here, but this documentation here is certainly better than not having it.

### DIFF
--- a/src/sssom_schema/schema/sssom_schema.yaml
+++ b/src/sssom_schema/schema/sssom_schema.yaml
@@ -64,7 +64,10 @@ enums:
 types:
  EntityReference:
     typeof: uriorcurie
-    description: A reference to a mapped entity. This is represented internally as a string, and as a resource in RDF
+    description: |
+      A reference to an entity involved in the mapping.
+      In the JSON and TSV serialisation, an entity reference MUST be a CURIE.
+      In the RDF-based serialisations, an entity reference MUST be an IRI/URI.
     base: str
     uri: rdfs:Resource
 
--- a/src/sssom_schema/schema/sssom_schema.yaml
+++ b/src/sssom_schema/schema/sssom_schema.yaml
@@ -66,10 +66,10 @@ types:
     typeof: uriorcurie
     description: |
       A reference to an entity involved in the mapping.
-      In the JSON and TSV serialisation, an entity reference MUST be a CURIE.
-      In the RDF-based serialisations, an entity reference MUST be an IRI/URI.
     base: str
     uri: rdfs:Resource
+    see_also:
+      - https://mapping-commons.github.io/sssom/spec/#tsv
 
 slots:
   mirror_from:


gouttegd (changes requested):
Specific details about the serialisations do not belong in the data model. The data model is (or must be) independent of any serialisation format.

In the model, an entity reference is an identifier. That’s it.

If a serialisation format wants to impose a specific format for the identifiers (e.g., in SSSOM/TSV, we only want CURIEs), that must be documented in the specification of that serialisation format.

Of course the problem right now is that we do not have proper specifications for the serialisation formats, so we have no place where to describe those details. That’s part of what I want to address in #330.

matentzn:
Ok, lets replace this PR by a "see also" reference to the doc on serialisations.

That said, the current description of Entity reference is still terrible :P I will leave this in draft mode to fix it another time.

Thanks for weighing in!

gouttegd:
Note that I can’t say when I will complete the reorganisation of the docs, so if you don’t want to wait for that, you can add a note to the [TSV section](https://mapping-commons.github.io/sssom/spec/#tsv) of the overview (which is the best/only thing we have as a “specification” for the TSV format, for now) to explicitly state that identifiers in a SSSOM/TSV file MUST be in CURIE form.

Something like

> All identifiers in a SSSOM/TSV file (both in the metadata section and in the TSV columns) MUST be in CURIE form. The use of full-length identifiers is officially not supported.

> That said, the current description of Entity reference is still terrible.

Agreed. “A reference to a mapped entity” is plain wrong, actually, since `EntityReference` is not only used for the `subject_id` and `object_id`, but for all cases where an identifier is required -- including identifiers to other things than the entities that are being mapped.

joeflack4:
I think I agree @gouttegd that the documentation for the model should be serialization agnostic. However, from a user experience perspective, I worry about the situation where someone goes to the documentation looking for an answer and they wind up looking at the documentation for the LinkML data model rather than something that is more relevant to the user; the serialization.

So I think I'll add a suggestion to your pull request, or it can be in a new PR, to update the documentation such that the docs for the serialization formats are front and center instead. And then I wonder if the LinkML documentation should be removed, or I suppose better yet, put aside in some developer documentation subsection. 

I'm also fine with adding a "see also" linking to the serialization documentation, but with the changes I suggested above, that is less of a necessary / useful change than it otherwise would be.

gouttegd:
> I worry about the situation where someone goes to the documentation looking for an answer and they wind up looking at the documentation for the LinkML data model rather than something that is more relevant to the user

That’s definitely something that is likely to happen right now. In fact the documentation for the data model is currently the _home page_ for the [SSSOM website](https://mapping-commons.github.io/sssom/), which is very unfortunate and is one of the things I want to change (changes are underway in [that branch](https://github.com/mapping-commons/sssom/tree/reorganise-docs).

> to update the documentation such that the docs for the serialization formats are front and center instead. And then I wonder if the LinkML documentation should be removed, or I suppose better yet, put aside in some developer documentation subsection.

That’s the plan.

gouttegd:
By the way, among the small things that need to be updated: [the examples in the spec](https://mapping-commons.github.io/sssom/spec/#tsv) consistently use a full-length IRI for the `creator_id` slot, despite that slot being a `EntityReference`:

```
#creator_id: "https://orcid.org/0000-0002-7356-1779"
#curie_map: 
#  HP: "http://purl.obolibrary.org/obo/HP_"
#  MP: "http://purl.obolibrary.org/obo/MP_"
#  skos: "http://www.w3.org/2004/02/skos/core"
#license: "https://creativecommons.org/publicdomain/zero/1.0/"
#mapping_provider: "http://purl.obolibrary.org/obo/upheno.owl"
subject_id  predicate_id    object_id   mapping_justification   subject_label   object_label
HP:0009124  skos:exactMatch MP:0000003  Lexical Abnormal adipose tissue morphology  abnormal adipose tissue morphology
HP:0008551  skos:exactMatch MP:0000018  Lexical Microtia    small ears
HP:0000411  skos:exactMatch MP:0000021  Lexical Protruding ear  prominent ears
```

It’s bad enough for the spec never to explicitly say that full-length IRIs are not expected in the TSV format, but giving examples where full-length IRIs are actually used is very bad.

src/sssom_schema/schema/sssom_schema.yaml
```diff
@@ -64,9 +64,12 @@ enums:
 types:
  EntityReference:
     typeof: uriorcurie
-    description: A reference to a mapped entity. This is represented internally as a string, and as a resource in RDF
+    description: |
+      A reference to an entity involved in the mapping.
     base: str
     uri: rdfs:Resource
+    see_also:
+      - https://mapping-commons.github.io/sssom/spec/#tsv
```
gouttegd, line 72: Just to note that this link _may_ need to change in the future (not sure yet, but after the overhaul the specification for the TSV format will likely be in a dedicated file, rather than in a section inside another file) – but we can change it later if/when needed.

ehartley (approved):
These changes look good when combined with the changes in #362.

matentzn:
@ehartley @gouttegd thank you for your reviews.

Issue #305, opened 2023-07-25
Clarify meaning of slots that exist both at mapping-level and at mapping set-level


gouttegd:
The mapping set class seems to have two very different types of metadata slots:

* Slots that are really for metadata of the _set_ itself, not of the mappings;
* Slots that are really for metadata of the individual mappings, but that are allowed to be used once and for all at the level of the mapping set – presumably to avoid repeating the same metadata for all mappings, if said metadata is the same for all individual mappings in a set.

It is obvious that the metadata slots that are only allowed at the set level belong to the first category (e.g., `mapping_set_id`, `mapping_set_title`).

But for the slots that are allowed both at the set level and at the individual mapping level, it is _not_ obvious how they should be interpreted when they are found at the set level.

If a set has a `creator_id` metadata for example, should that be understood to mean that the referenced creator is the creator of the _set_ (which may not be the same as the creator(s) of the individual mappings that make up the set), or that it is the default creator of all the mappings of the set (at least all those that do not have an explicit value for that slot)?

The documentation of `creator_id` seems to suggest the latter because it refers to “mapping” in singular (“Identifies the persons or groups responsible for the creation of the mapping“), but I’d be wary of using English grammar to infer whether a metadata slot is intended to refer to an individual mapping or to a set of mappings… especially since the documentation of `author_id`, which is a slot that only exists for individual mappings (and so cannot refer to the “author” of a mapping _set_) uses “mappings” in plural!

For now, my assumption is that _all_ the slots that exist both at the mapping level and at the mapping set level are all about the individual mapping, and that when they are used at the set level, it is _only_ as a “shortcut” to avoid repeating the same metadata for all individual mappings (said otherwise: a slot never has a different meaning depending on whether it is used on an individual mapping or on a mapping set). But this is something that should be explicitly specified. All the spec says is:

> Some SSSOM metadata elements apply only to one element or the other, but many can be applied to both.

which in my opinion is wholly inadequate.

gouttegd:
To further clarify: the issue above matters because it dictates whether metadata found at the set level should be propagated down to the mapping level, or not.

If setting `creator_id` at the set level is merely a shortcut to avoid explicitly setting `creator_id` on each mapping (as I’m currently assuming it is), then it is fine to propagate the value of that slot to all the mappings of the set when de-serializing a set – in fact, not only it is fine, but arguably that is what _should_ be done.

[...]

PR #371, opened 2024-07-11
Model propagatable slots

gouttegd:
Resolves [#305]

- [ ] provide a full, working and valid example in `examples/` (**not applicable**: no new example needed as the change only affects how some slots should be interpreted; it does not add or remove slots, nor does it change how the propagated slots are used)
- [ ] provide a link to a valid example in the `see_also` field of the linkml model (**not applicable**, same reason as above)

This PR finalises the fix to #305, by explicitly specifying, directly within the LinkML model, which slots are considered “propagatable” (previously this was only informally described in the spec, since #368). This is done by:

* adding a “metamodel extension class“ (`sssom:Propagatable`) with a single boolean-ranged attributed `propagated`;
* amending the slots that must be considered propagatable by making them instantiate the `sssom:Propagatable` extension.

### DIFF
--- a/src/sssom_schema/schema/sssom_schema.yaml
+++ b/src/sssom_schema/schema/sssom_schema.yaml
@@ -173,6 +173,9 @@ slots:
   subject_type:
     description: The type of entity that is being mapped.
     range: entity_type_enum
+    instantiates: sssom:Propagatable
+    annotations:
+      propagated: true
     examples:
       - value: owl:Class
   predicate_id:
@@ -283,6 +286,9 @@ slots:
   object_type:
     description: The type of entity that is being mapped.
     range: entity_type_enum
+    instantiates: sssom:Propagatable
+    annotations:
+      propagated: true
     examples:
       - value: owl:Class
   mapping_set_id:
@@ -363,6 +369,9 @@ slots:
   subject_source:
     description: URI of vocabulary or identifier source for the subject.
     range: EntityReference
+    instantiates: sssom:Propagatable
+    annotations:
+      propagated: true
     examples:
       - value: obo:mondo.owl
         description: A persistent OBO CURIE pointing to the latest version of the Mondo ontology.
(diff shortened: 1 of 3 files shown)


hrshdhgd (approved):
Looks good to me

src/sssom_schema/schema/sssom_schema.yaml
matentzn, line 1: Everything looks good, but I one clarifying question: Does any of the changes you are proposing to the model affect the RDF or JSON (LD) serialisation of SSSOM (since its using the LinkML dumper classes)?

hrshdhgd: Ooh ... that is a great question!! I hadn't thought of that aspect ...good catch!

gouttegd: I see no reason why the RDF serialisation would change in the slightest. The change here is the addition of an annotation to the meta model, not the data represented by the model. And the RDF serialisation produces no traces at all of the meta model.

As for the JSON(LD) serialisation, it does produce a representation of the meta model in its output (in the JSON-LD `@context` attribute), alongside the actual data:

```json
{
  "mapping_set_id": "http://purl.obolibrary.org/obo/fbbt/fbbt-mappings.sssom.tsv",
  "license": "https://creativecommons.org/licenses/by/4.0/",
[... TRUNCATED FOR BREVITY ...]
  "@type": "MappingSet",
  "@context": {
    "author_id": {
      "@type": "rdfs:Resource",
      "@id": "pav:authoredBy"
    },
    "comment": {
      "@id": "rdfs:comment"
    },
    "confidence": {
      "@type": "xsd:double"
    }
[... TRUNCATED FOR BREVITY ...]
}
```

so I would expect that the `propagated` annotation would appear in the `@context`'s description of the propagatable slots, in some form or another.

However, I am not overly concerned about that, since

1) The JSON-LD output is completely unspecified, so _no one_ should have workflows that are dependent on it -- especially not workflows that could break just because one new key suddenly appears in the `@context`.

2) The JSON-LD output is, well, _crap_, which is another reason why no one should depend on it. I’ve mentioned that before, but it bears being mentioned again: what the hell is the description of some the slots even supposed to mean:

```json
"mapping_cardinality": {
  "@context": {
    "@vocab": "@null",
    "text": "skos:notation",
    "description": "skos:prefLabel",
    "meaning": "@id"
  }
}
```

matentzn: That’s what I suspected; in this case, good to go

matentzn (approved):
Looks good. Thank you!

PR #373, opened 2024-07-18
Add min/max value constraints to score and confidence slots.

### LINKED ISSUE #372: Range constraints for Double-typed slots not specified/enforced
The `confidence`, `semantic_similarity_score`, and `similarity_score` slots are all described as accepting a value between 0 and 1, but this constraint is not formalised in the model. LinkML does provide [a way to describe such constraints](https://linkml.io/linkml/schemas/constraints.html#minimum-and-maximum-values) (`minimal_value` / `maximal_value`), so we should make use of it. Presumably the LinkML runtime (at least for Python implementations) could then automatically enforce those constraints and flag mapping sets that are using out-of-range values.

gouttegd:
Resolves #372

- [ ] `docs/` have been added/updated if necessary (not applicable; the doc was already ahead of the model, this PR is about updating the model to align it with the doc)

The description of the `confidence`, `semantic_similarity_score`, and `similarity_score` slots says that those slots expect a value between 0 and 1, but the model does not enforce that. LinkML provides a way to formalise this kind of value constraints (which could then be automatically enforced by a runtime that is aware of those constraints), so we do exactly that here: we add `minimum_value` and `maximum_value` constraints to the aforementioned slots.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 - Add issue_tracker_item and issue_tracker [model elements](https://github.com/mapping-commons/sssom/pull/259).
 - Add recommendation to sort the keys in the YAML metadata block.
+- Double-typed slots explicitly constrained to the [0.0,1.0] range, as per their description.
 
 ## SSSOM version 0.13.0
 
--- a/src/sssom_schema/schema/sssom_schema.yaml
+++ b/src/sssom_schema/schema/sssom_schema.yaml
@@ -449,6 +449,8 @@ slots:
     description: A score between 0 and 1 to denote the confidence or probability that
       the match is correct, where 1 denotes total confidence.
     range: double
+    minimum_value: 0.0
+    maximum_value: 1.0
   subject_match_field:
     description: A list of properties (term annotations on the subject) that was used
       for the match.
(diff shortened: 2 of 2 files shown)


matentzn (approved):
We probably need a better SOP for testing these kinds of PR. I had been working on a pipeline, but the truth is, what would actually need to happen is:

1. Build LinkML model (this ensures that the model is syntactically well-defined)
2. Build version of sssom py with that locally cached linkml model
3. Run sssom-py test suites

Else I dont know if a change like this will break existing tests. Did you test these changes somehow?

gouttegd:
> 1. Build LinkML model (this ensures that the model is syntactically well-defined)

I believe the existing test suite here (the one that is run by `make test`) takes care of validating the model at least from a syntactical standpoint. First, because what would even be the point of the test suite if it did not even do that, and second, because when preparing this PR I initially and erroneously wrote `maximal_value` instead of `maximum_value`, and the test suite (which I ran locally before even committing any changes) correctly rebuked me for that.

> 2. Build version of sssom py with that locally cached linkml model
> 3. Run sssom-py test suites

Why would we want `sssom` to fail because of problems that could exist in `sssom-py`?

If SSSOM-Py breaks when the model enforces a constraint that has always been a part of the spec but was not previously enforced, then it's a problem with SSSOM-Py, not with the spec.

matentzn:
> I believe the existing test suite here (the one that is run by make test) takes care of validating the model at least from a syntactical standpoint.

You are right, forgot about that; it already does what I thought it did not.

hrshdhgd (approved):

PR #375, opened 2024-07-23
Add support for extension slots

### LINKED ISSUE #328: Dealing with non-standard slots
Among the issues that I believe should be decided along the road to 1.0 is: How should implementations deal with the possible presence of non-standard slots (slots not described anywhere in the specification and/or the schema) in a mapping set?

That is, considering this mapping set (note the presence of a non-standard _foo_ slot in the set metadata and a non-standard _bar_ and _baz_ slots in the mappings metadata.):

```
#curie_map:
#  FBbt: "http://purl.obolibrary.org/obo/FBbt_"
#  UBERON: "http://purl.obolibrary.org/obo/UBERON_"
#foo: "The yellow fox jumped over the lazy dog"
subject_id      predicate_id    object_id       mapping_justification   bar     baz
FBbt:00000001   semapv:crossSpeciesExactMatch   UBERON:0000468  semapv:ManualMappingCuration    A       D
FBbt:00000002   semapv:crossSpeciesExactMatch   UBERON:6000002  semapv:ManualMappingCuration    B       E
FBbt:00000003   semapv:crossSpeciesExactMatch   UBERON:0000914  semapv:ManualMappingCuration    C       F
```

what should implementations do?

The current behaviour of both `sssom-py` and `sssom-java` is to _ignore_ the non-standard slots (silently, in the case of `sssom-java`; `sssom-py` emits a warning) and to accept the rest of the set. That is, running this set through (for example)

```sh
sssom parse -I tsv sample.sssom.tsv -o output.sssom.tsv
```

will produce

```
#curie_map:
#  FBbt: "http://purl.obolibrary.org/obo/FBbt_"
#  UBERON: "http://purl.obolibrary.org/obo/UBERON_"
subject_id      predicate_id    object_id       mapping_justification
FBbt:00000001   semapv:crossSpeciesExactMatch   UBERON:0000468  semapv:ManualMappingCuration
FBbt:00000002   semapv:crossSpeciesExactMatch   UBERON:6000002  semapv:ManualMappingCuration
FBbt:00000003   semapv:crossSpeciesExactMatch   UBERON:0000914  semapv:ManualMappingCuration
```

I think this behaviour is sane and should be explicitly prescribed in the specification (“Implementations SHOULD ignore any non-standard slot not described in this specification; they MAY print a warning upon encountering any such slot.”).

A related issue is: If a SSSOM producer does want to include additional metadata for which there is no suitable standard slots, and wants those additional metadata to be preserved by the existing tooling (instead of being ignored and discarded as above), what should be the proper way to do that?

Currently, the official way is through the (poorly specified) `other` slot, which accepts a “list of key value pairs for properties not part of the SSSOM spec” and “can be used to encode additional provenance data”. So, according to the current spec, the example above should have been:

```
#curie_map:
#  FBbt: "http://purl.obolibrary.org/obo/FBbt_"
#  UBERON: "http://purl.obolibrary.org/obo/UBERON_"
#other:
#  - "foo=The yellow fox jumped over the lazy dog"
subject_id      predicate_id    object_id       mapping_justification   other
FBbt:00000001   semapv:crossSpeciesExactMatch   UBERON:0000468  semapv:ManualMappingCuration    bar=A|baz=D
FBbt:00000002   semapv:crossSpeciesExactMatch   UBERON:6000002  semapv:ManualMappingCuration    bar=B|baz=E
FBbt:00000003   semapv:crossSpeciesExactMatch   UBERON:0000914  semapv:ManualMappingCuration    bar=C|baz=F
```

This method of encoding additional metadata doesn’t look great, which is probably why there is [a proposition to replace it](https://github.com/mapping-commons/sssom/pull/263), in which additional slots would be explicitly defined in a `extension_definition` slot:

```
#extension_definition:
#  - key: bar
#    value: <type of value expected in foo>
#  - key: baz
#    value: <type of value expected in baz>
```

I think that whatever method we think is best to encode additional metadata, it should be decided in time for 1.0.

I’d like to propose a simpler method than the one drafted in #263. I like the idea of explicitly declaring the additional metadata but I don’t see the point of specifying the type of value. Non-standard metadata would be tied to a particular use-case and to particular users, so it should be up to them to know what to expect in any non-standard field that they need.

Instead, I propose that non-standard slots should simply be declared as a list of slot names:

```
#curie_map:
#  FBbt: "http://purl.obolibrary.org/obo/FBbt_"
#  UBERON: "http://purl.obolibrary.org/obo/UBERON_"
#extra_set_slots:
#  - foo
#extra_slots:
#  - bar
#  - baz
#foo: "The yellow fox jumped over the lazy dog"
subject_id      predicate_id    object_id       mapping_justification   bar     baz
FBbt:00000001   semapv:crossSpeciesExactMatch   UBERON:0000468  semapv:ManualMappingCuration    A       D
FBbt:00000002   semapv:crossSpeciesExactMatch   UBERON:6000002  semapv:ManualMappingCuration    B       E
FBbt:00000003   semapv:crossSpeciesExactMatch   UBERON:0000914  semapv:ManualMappingCuration    C       F
```

The effect of such a declaration would be that the declared slots should be _preserved exactly as they are_, instead of being _ignored_. That is, the mapping set above should be “round-trippable” through `sssom-py parse`, because the `foo`, `bar`, and `baz` slots, being declared, should not be discarded.

Thoughts?

gouttegd:
Resolves #328

This PR updates the SSSOM model to allow for _defined extensions_ as proposed in #328.

It also updates the description of the data model to describe the use of both defined and undefined extension, and the specification of the SSSOM/TSV format to explain how SSSOM/TSV parsers and writers should deal with such extensions.

Overall, this is exactly what was proposed in [this comment](https://github.com/mapping-commons/sssom/issues/328#issuecomment-1879735143) in #328, except that here we need to split the specification in two parts (one about extensions in general, independently of the serialisation format, and one about the SSSOM/TSV serialisation of extensions), while the initial proposition was in a single block.


src/docs/spec-formats-tsv.md
```diff
@@ -208,6 +230,13 @@ When writing the metadata block, a canonical SSSOM/TSV writer:
 * MUST NOT include in the CURIE map any prefix name that is not used anywhere in the set;
 * MUST sort the prefix names in the CURIE map in lexicographical order.
 
+In addition, if [extension slots](spec-model.md#non-standard-slots) are supported, the writer:
+
+* MUST write any extension slot in the mapping set _after_ the standard slots;
+* MUST sort the extension slots lexicographically on the `property` of their corresponding extension definitions;
+* MUST sort extension definitions on their `property` value;
+* MUST not include an extension definition if the corresponding extension is not used anywhere in the set.
```
matentzn, line 238: NON-ACTIONABLE COMMENT: I am a bit ambivalent about this one, similar to the question of rendering empty columns, as there is a use case for serialising mapping sets that are "work in progress", i.e. you pipe a SSSOM TSV through one process (say "filter) and then publish it so it can be edited my a human workflow. I just want to put this thought out there, not block this specific restriction as the user can always use the flag for not canonicalising if they have this specific need.

src/docs/spec-model.md
```diff
@@ -81,3 +81,66 @@ In addition, predicates from the following sources MAY also be encouraged:
 
 * any relation from the [Relation Ontology (RO)](https://obofoundry.org/ontology/ro.html);
 * any relation under [skos:mappingRelation](http://www.w3.org/2004/02/skos/core#mappingRelation) in the [Semantic Mapping Vocabulary](https://mapping-commons.github.io/semantic-mapping-vocabulary/).
+
+## Non-standard slots
+
+<a id="non-standard-slots"></a>
+
+Implementations are only REQUIRED to support the standard metadata slots defined in the LinkML model.
```
matentzn, line 89: > defined in the SSSOM LinkML schema

So to not cause confusion that we re talking about the sssom schema here (rather than the LinkML meta-schema)

src/docs/spec-model.md
```diff
@@ -81,3 +81,66 @@ In addition, predicates from the following sources MAY also be encouraged:
...
+
+### Restrictions on the values of extension slots
+
+#### General restrictions
+
+The following restrictions apply to all extension slots, regardless of whether they are defined or undefined.
+
+Each mapping set and each mapping can have at most _one_ value for each extension slot. The expected behaviour upon encountering a repeated extension slot is unspecified.
+
+All extension values MUST be representable as literal strings. Complex data structures (e.g., lists or dictionaries) MUST NOT be used.
```
matentzn, line 132: Note: _all_ data structures can be represented as a literal string, so not sure if this sentence will do anything. I can have an extension slot, type it as xsd:string, and shove a 10 GB XML document in there if I so want to..

Another way to put it probably is to say that the expected type should be a simple, primitive type, and not a complex, composite type (like list or dict or some such).

matentzn (commented):
I love it. Made some comments, have not done any testing yet.

examples/schema/extension-slots.sssom.tsv
```diff
@@ -0,0 +1,23 @@
...
+#mapping_set_id: https://example.org/sets/exo2c-with-extensions
+#mapping_set_title: Sample set EXO2C with extension slots
+#license: https://creativecommons.org/licenses/by/4.0/
+#extension_definitions:
+#  - slot_name: ext_bar
+#    property: EXPROP:barProperty
+#    type_hint: xsd:integer
+#  - slot_name: ext_baz
+#    property: EXPROP:bazProperty
+#    type_hint: linkml:Uriorcurie
```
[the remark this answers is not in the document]
gouttegd: The `linkml` prefix is made a built-in prefix in this very PR (see the changes to the table at the bottom of `spec-intro.md`).

This is as proposed in point 2.g [here](https://github.com/mapping-commons/sssom/issues/328#issuecomment-1879735143):

> For convenience, the following two prefix names are added to the list of the built-in prefixes that do not need to be declared in the curie_map:
>
> * `xsd: http://www.w3.org/2001/XMLSchema#`,
> * `linkml: https://w3id.org/linkml/`.

src/docs/spec-model.md
gouttegd: I wonder how could someone interpret “the LinkML model” in this context as referring to anything else than “the model defined by the LinkML schema we are talking about since the beginning of this document”, but fine, if you believe there is a possible ambiguity here, happy to clear it out.

src/docs/spec-model.md
gouttegd: > Another way to put it probably is to say that the expected type should be a simple, primitive type, and not a complex, composite type (like list or dict or some such).

Agreed.

> I can have an extension slot, type it as xsd:string, and shove a 10 GB XML document in there if I so want to..

Note that as long as it is encoded as a string (as in `my_extension_slot: '<?xml version="1.0"?><article xmlns="http://docbook.org/ns/docbook/">...</article>'), that shouldn’t be a problem at all.

src/docs/spec-model.md
matentzn: Thanks for accomodating, this was my impression on my first read.

src/docs/spec-model.md
matentzn: Your new docs look perfect.

src/docs/spec-formats-tsv.md
```diff
@@ -134,6 +134,28 @@ For any given propagatable slot, condensation is only allowed if (1) all mapping
 Implementations that support condensation MUST also support propagation.
 
 
+## Non-standard slots
+
+If an implementation does not support [non-standard slots](spec-model.md#non-standard-slots), then:
+
+* a SSSOM/TSV reader MUST discard any unknown top-level YAML key in the metadata block, and any unknown TSV column in the TSV section;
+* a SSSOM/TSV reader MUST NOT write any unknown top-level YAML key in the metadata block, or any unknown TSV column in the TSV section.
```
matentzn, line 142: - "reader" ... MUST NOT .... "write"
- Also, seems redundant with the above?

src/docs/spec-model.md
```diff
@@ -81,3 +81,68 @@ In addition, predicates from the following sources MAY also be encouraged:
...
+
+* the name of the slot, as it will appear when used in a mapping set (`slot_name`);
+* a property intended to specify the meaning of the slot (`property`);
+* the type of values expected by the slot (`type_hint`).
+
+A definition MUST have at least a `slot_name`. The name MUST be a XML “non-colonized name” (“NCName”, see [Namespaces in XML, §2](https://www.w3.org/TR/1999/REC-xml-names-19990114/#NT-NCName)). The name MUST NOT match the name of an existing standard slot.
+
+To avoid any conflicy with a future version of the SSSOM specification (which could introduce new standard slot names), implementations are strongly encouraged to craft extension slot names that start with the `ext_` prefix. No new standard slot with a name starting with `ext_` will ever be introduced in any future version of the standard. (This is an advice for SSSOM producers only; SSSOM consumers MUST NOT reject an extension slot solely on the basis that its name does not start with `ext`.)
+
+A definition SHOULD have a `property`. If it does not, implementations MUST automatically construct a default property by catenating the prefix `http://sssom.invalid/` with the name of the extension.
```
matentzn, line 107:
```suggestion
A definition SHOULD have a `property`. If it does not, implementations MUST automatically construct a default property by concatenating the prefix `http://sssom.ext/` with the name of the extension.
```

Why invalid?

src/docs/spec-model.md
```diff
@@ -81,3 +81,68 @@ In addition, predicates from the following sources MAY also be encouraged:
...
+
+The slot name and the property MUST be unique to each definition. No two definitions can share the same name and/or the same property.
```
matentzn, line 109: > No two definitions can share the same name and/or the same property.

Why property? Typical example:

- nico_comment
- damien_comment
- harshad_comment

all have the rdfs:comment property?

matentzn: (As an aside, while I think this part of the definition should be dropped, there is a bit of a risk that the property slot in the extensions clashes with the "meaning" or "slot_uri" values in the regular sssom slots.

src/docs/spec-model.md
```diff
@@ -81,3 +81,68 @@ In addition, predicates from the following sources MAY also be encouraged:
...
+
+Once defined, an extension slot may be used as a supplementary slot in either the `Mapping` class or the `MappingSet` class (or both), as if it was a normal, standard slot. How those slots are represented internally and provided to client code is left at the discretion of the implementations.
+
+### Undefined extensions
+
+Undefined extensions are non-standard slots that are not explicitly defined as described in the previous section. Implementations MAY support undefined extensions.
+
+Upon encountering a non-standard slot that is not a defined extension, an implementation that supports undefined extensions MUST behave as if the slot had been defined with:
+
+* a `property` constructed by catenating the prefix `http://sssom.invalid/` to the name of the slot;
```
matentzn, line 121: If you correct the above (i dont like "invalid" at all here, something more neutral would be preferable), remember to correct it hear as well

matentzn (commented):
Some last round of comments, I looked how the generated schemas look like and I am satisfied with the formal representation. The most important part of this PR is in any case the update to the specification.

src/docs/spec-formats-tsv.md
gouttegd: The second “reader” is indeed a mistake, I’ll fix it.

As for the redundancy, I don’t think so. The spec don’t say anywhere else what implementations are supposed to do with unknown slots.

src/docs/spec-model.md
gouttegd: Because that’s an officially recognised top-level domain specifically intended for this kind of purposes ([RFC 2606](https://datatracker.ietf.org/doc/html/rfc2606)), one that is guaranteed never to resolve to anything ([RFC 6761](https://datatracker.ietf.org/doc/html/rfc6761)).

matentzn: Awesome! Didnt know that, and its fine by me then.

src/docs/spec-formats-tsv.md
matentzn: (No need to belabour the point, I thought "if the reader MUST discard, then the "writer" cant really have anything to write anymore - but I guess it does not hurt to be clear)

src/docs/spec-model.md
gouttegd: > Why property? Typical example:
>
>    nico_comment
>    damien_comment
>    harshad_comment
>
> all have the rdfs:comment property?

No! The _property_, not the *slot_name*, is what identifies the extension value. That’s why it is strongly recommended that every extension definition has a _property_, and why implementations MUST construct one automatically if needed. The slot name is merely how the extension is referred to in the SSSOM/TSV format.

You’re the one who insisted on that because it was critical for you that extension values could be serialised into RDF, which required them to have a proper IRI instead of merely a name. How would you serialise your “typical example” in RDF if your three extension slots all have the same `rdfs:comment` property???

matentzn: Ok I dont remember all the details on this discussion, I apologise. I was happy with our final draft so I wont change my mind now!

src/docs/spec-formats-tsv.md
gouttegd: > if the reader MUST discard, then the "writer" cant really have anything to write anymore

A writer can have to write a set that is _not_ the result of a prior parsing operation (e.g. if the set has been built from scratch by the application).

matentzn: Of course!

src/docs/spec-formats-tsv.md
matentzn: Resolving this.

matentzn (dismissed):
Awesome, done now, getting other reviewers in!

hrshdhgd (changes requested):
Same [comment](https://github.com/mapping-commons/sssom/pull/386#pullrequestreview-2217867625) as #386

Any change to the YAML file should be accompanied by a `make gen-project` to update the corresponding python file.

hrshdhgd (approved):

matentzn (approved):

matentzn:
Ready to merge!

gouttegd:
To re-reviewers: the only change was to the changelog, to fix the merge conflict caused by the merge of #386.

This is the expected consequence of the requirement (stated in the PR template) that the changelog must be updated in every PR (instead of updating it in one go before a release). All additions to the changelog are done at exactly the same place, so when there are several concurrent PR there cannot not be a merge conflict.

gouttegd:
Will do by this evening. :)

matentzn (approved):
awesome feature!

hrshdhgd (approved):

PR #349, opened 2024-02-06
Add curie_map element to data model

matentzn:
Resolves #225 

- [x] `docs/` have been added/updated if necessary
- [X] `make test` has been run locally
- [x] tests have been added/updated (if applicable)
- [x] [CHANGELOG.md](https://github.com/mapping-commons/sssom/blob/master/CHANGELOG.md) has been updated.

If you are proposing a change to the SSSOM metadata model, you must 

- [x] provide a full, working and valid example in `examples/`
- [x] provide a link to the related GitHub issue in the `see_also` field of the linkml model
- [x] provide a link to a valid example in the `see_also` field of the linkml model
- [x] Update the spec to reflect the fact that the curie_map is now part of the model

This PR adds the `curie_map` as an official property to the SSSOM data model.

For testing this, I followed @gouttegd [great docs](https://github.com/mapping-commons/sssom/blob/master/CONTRIBUTING.md#local-testing). But I also ran some of my own, local tests to ensure that sssom-py does not break.

PR #376, opened 2024-07-24
Add `curie_map` to the model

### LINKED ISSUE #225: [New metadata element]: `MappingSet.curie_map`
`curie_map`

```
curie_map:
    OMIM: http://www.omim.org/something/something
```

```
map
```

**Description**
A map of CURIEs. CURIE maps already exist in the SSSOM metadata and many files that are currently in use, so strange for it not to be in the spec. It is also commonly called `prefix_map`.

gouttegd:
Resolves #225

- [ ] tests have been added/updated (not applicable – nothing new to test, the CURIE map has existed for a while, this PR just formalises it directly into the model)

- [x] provide a full, working and valid example in `examples/` (no need for a new example – all existing examples already have a `curie_map`)
- [ ] provide a link to a valid example in the `see_also` field of the linkml model (no need, see above)

This PR makes the `curie_map` a _bona fide_ slot of the `MappingSet` class, that exists independently of the serialisation format.

It builds on top of #349 by explaining in the text version of the spec what the `curie_map` slot is for, and how it should be dealt with in the SSSOM/TSV serialisation.


src/docs/spec-formats-tsv.md
```diff
@@ -203,7 +201,6 @@ When writing the metadata block, a canonical SSSOM/TSV writer:
 * MUST serialise multi-valued slots as YAML “block sequences” ([YAML Specification §8.2.1](https://yaml.org/spec/1.2.2/#821-block-sequences)) – even when the list of values contains only one item;
 * MUST serialise scalar values in YAML “plain style” ([YAML Specification §7.3.3](https://yaml.org/spec/1.2.2/#733-plain-style)) whenever possible, otherwise in “double-quoted style” ([YAML Specification §7.3.1](https://yaml.org/spec/1.2.2/#731-double-quoted-style));
 * MUST serialise the slots in the order they appear in the [“Slots” table](MappingSet.md#slots), in the documentation for the `MappingSet` class;
-* MUST write the `curie_map` at the beginning of the block, before any other slots;
```
matentzn, line 206: I dont think we can drop this, as the element is not tagged "required" in the spec. The fact that it is written at the beginning is redundant, yes, but not the fact that it must be written at all.

matentzn (commented):
I added the "example" after all, just for completeness reasons. One small comment, then we can merge.

matentzn:
_Awesome_ thank you!!!

src/docs/spec-formats-tsv.md
gouttegd: The fact that the `curie_map` must be written at all is a general requirement that applies to all SSSOM/TSV files, not only to the ”canonical” variant. And that requirement is already stated elsewhere (“Identifiers” section):

> As stated in the description of the model, all prefix names used in CURIEs MUST be declared in the `curie_map` slot of the mapping set object

which obviously implies that the `curie_map` slot itself MUST be present, even if it is not marked as “required” in the LinkML schema.

So the removed line is completely redundant.

matentzn: Yep makes sense. Thanks!

matentzn (approved):
Looks great, thank you!

examples/schema/curie_map.sssom.tsv
```diff
@@ -0,0 +1,13 @@
+#curie_map:
+#  HP: http://purl.obolibrary.org/obo/HP_
+#  MP: http://purl.obolibrary.org/obo/MP_
+#  orcid: https://orcid.org/
+#mapping_set_id: https://w3id.org/sssom/commons/examples/curie_map.sssom.tsv
+#license: "https://creativecommons.org/publicdomain/zero/1.0/"
+#creator_id: orcid:0000-0002-7356-1779
```
gouttegd, line 7: Please do not create examples that represents multi-valued slots as if they were singled-valued slots. `creator_id` expects a list; even if the list has one value, it should be represented as a proper list:

```
#creator_id:
```

Yes, we have [explicitly allowed](https://mapping-commons.github.io/sssom/spec-formats-tsv/#multi-valued-slots-with-a-single-value) the use of the single-value syntax, but the spec is clear that this was only done for compatibility with existing sets only and that multi-valued slots should otherwise be represented as real lists. We should not provide bad examples like that.

matentzn (dismissed):

hrshdhgd (changes requested):
Sorry to repeat this in every PR

run `make all` same as https://github.com/mapping-commons/sssom/pull/386#pullrequestreview-2217867625

gouttegd:
> Sorry to repeat this in every PR

No need to be sorry. However:

1. Re-generating and committing all derived files upon every PR is, I believe, a very bad idea.

* It makes even the slightest PR much larger than it needs to be, and therefore more cumbersome to review.
* It pollutes the history of the project with large changes that do not mean anything (only the changes to the source schema are actually meaningful).
* It greatly increases the likelihood of merge conflicts when there are several PRs in the works.

It would be much better in my opinion to leave the auto-generated files untouched most of the times, and only commit them once in a while (typically shortly after a release).

2. If you insist on having the derived files re-generated upon every PR, then it‘s something that should be mentioned in the `CONTRIBUTING.md` file, which for now has no trace at all of such of requirement – instead it’s full of useless, generic guidelines that were mindlessly copied over from another project.

And since there is a PR template, that requirement should also be mentioned in it.

hrshdhgd (approved):

gouttegd:
I‘ll fix the conflict here once #375 will have been (re-)approved ­– no point in fixing it now since merging #375 will (again) cause a conflict in the changelog.

matentzn (approved):
Looks great, thank you!

PR #377, opened 2024-07-31
Docs restructuring, adding funding and training pages

matentzn:
In this PR we restructure the documentation pages a little, and add pages that detail the main funding sources and training resources.

- [ ] ~~`make test` has been run locally~~


src/docs/funding.md
```diff
@@ -0,0 +1,51 @@
...
+
+The Simple Standard for Sharing Ontological Mappings (SSSOM) is a community-driven project which has received support from many different sources.
+We list the most important ones in the following.
+
+### Volunteering efforts
+
+A huge fraction of the work on SSSOM has been done by volunteers without dedicated grant support.
+We hereby acknowledge their contributions as being absolutely essential. A selection of amazing contributions (by no means exhaustive):
+
+- The development of [SSSOM Java](https://incenp.org/dvlpt/sssom-java/)
```
gouttegd, line 11: I am not sure this is necessary (and I would never have added that myself), but I’d be lying if I said that I do not appreciate it. Thanks!

src/docs/funding.md
```diff
@@ -0,0 +1,51 @@
...
+- Hundreds of careful contributions to discussions on the [SSSOM issue tracker](https://github.com/mapping-commons/sssom/issues)
+- The first draft of the [Mapping Registry Cookiecutter](https://github.com/mapping-commons/mapping-commons-cookiecutter)
+- We try to keep track of other [Community efforts here](https://github.com/mapping-commons/sssom/discussions/318)
+
+### Phenomics First (NIH / NHGRI #1RM1HG010860-01)
```
gouttegd, line 16: Please note that there is currently already a “funding” section at the bottom of the [home page](https://mapping-commons.github.io/sssom/#acknowledgements). I have no opinion on whether that information should be on the home page or on a separate page, but it shouldn’t be repeated in two different pages (if only to avoid possible discrepancies between the two pages).

I’d suggest removing the entire “Acknowledgements” section on the home page. Your call on whether you want to do that as part of this PR (no objection if you want to do that in a follow-up PR).

gouttegd (approved):
The only issue is the wrongly placed comment about `spec-model.md` needing rewriting (see below), but since this is a purely internal comment that is not publicly visible, no need to request changes for that (you might still want to update that though, if only to remember to re-write the correct page :D ).

src/docs/training.md
```diff
@@ -0,0 +1,27 @@
...
+
+- [Elevator pitch](#elevator)
+- [Tutorials and Guides](#guides)
+- [Related tutorials](#related)
+
+<a id="elevator"></a>
+
+### Elevator pitch
+
+<iframe width="560" height="315" src="https://www.youtube.com/embed/T-YP-Lv9k2s?si=tN9lHUFcovkyvm6z" title="YouTube video player" frameborder="0" allow="accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture; web-share" referrerpolicy="strict-origin-when-cross-origin" allowfullscreen></iframe>
```
gouttegd, line 11: A 11-minutes video? What kind of elevator are you taking? :D

matentzn: Hahah. You are right of course. There are some elevators in Athens that take 11 minutes to go 5 stories but that is not what the general public would understand :P

src/docs/funding.md
matentzn: I just replaced the section by a link rather than removing it altogether. Does this work?

src/docs/funding.md
matentzn: It stays!

mkdocs.yml
```diff
@@ -30,39 +30,38 @@ nav:
   - Specification:
     - Introduction: spec-intro.md
     - Data model:
-      - Introduction: spec-model.md
+      - Introduction: spec-model.md #This needs to be rewritten
+      - Applying Chaining Rules: chaining-rules.md
```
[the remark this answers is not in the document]
matentzn: Yes, this PR is only re-org, not overhaul!

src/docs/funding.md
gouttegd: It works for me.

gouttegd (approved):
Looks all good to me.

hrshdhgd (approved):

PR #381, opened 2024-08-03
Specify how to express an unmapped entity using sssom:NoTermFound

### LINKED ISSUE #28: How should unmapped elements be indicated?
I frequently want to be able to state: I have looked in `object_source`  for a matching or close concept to `subject_id` but am confident there is nothing. 

How should I state this?

Can we make object_id/name nullable (where we have a convention that '' in the TSV is null)?

alternatively we could have a CURIE/URI for NoMapping, analogous to owl:Thing.  

@wdduncan we need this for the nmdc sssom

I'm also using a pseudo-sssom for cob-to-external where I have blank entries for not mapped

matentzn:
Fixes #28

In the PR we add the ability to express that an entity in a mapping does not have a corresponding mappable entity.

In addition to the model entry, the feature is documented in the model specification.

As usual, we add examples and link to the corresponding issue.

- [ ] NA ~~tests have been added/updated (if applicable)~~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Next
 
 - Add the concept of "propagatable slots".
+- Add the entity reference `sssom:NoTermFound` of express the concept of an "unmapped entity" ([issue](https://github.com/mapping-commons/sssom/issues/28))
 
 ## SSSOM version 0.15.1
 
--- a/src/docs/spec-model.md
+++ b/src/docs/spec-model.md
@@ -81,3 +81,41 @@ In addition, predicates from the following sources MAY also be encouraged:
 
 * any relation from the [Relation Ontology (RO)](https://obofoundry.org/ontology/ro.html);
 * any relation under [skos:mappingRelation](http://www.w3.org/2004/02/skos/core#mappingRelation) in the [Semantic Mapping Vocabulary](https://mapping-commons.github.io/semantic-mapping-vocabulary/).
+
+## Documenting unmapped elements
+
+The absence of a mapping in a mapping set can be interpreted in two ways:
+
+1. There is no mapping for that specific (absent) mapping that can be established
+2. The mapping was not established for some other reason, for example, the mapping set is a work in progress
+
+To circumvent this ambiguity, we can explicitly state that a specific term cannot be mapped 
+by using a SSSOM-built-in entity called `sssom:NoTermFound`.
+
+Example SSSOM TSV file:
+
+```
+#curie_map:
+#  HP: http://purl.obolibrary.org/obo/HP_
+#  MP: http://purl.obolibrary.org/obo/MP_
+#  obo: http://purl.obolibrary.org/obo/
+#  orcid: https://orcid.org/
+#mapping_set_id: https://w3id.org/sssom/commons/examples/no_term_found.sssom.tsv
+#creator_id:
+#  - orcid:0000-0002-7356-1779
+#subject_source: obo:hp
+#object_source: obo:mp
+#license: "https://creativecommons.org/publicdomain/zero/1.0/"
+#mapping_provider: "https://w3id.org/sssom/core_team"
+#comment: This is an example file for the SSSOM for illustration only. Its contents are entirely fabricated.
+subject_id	predicate_id	object_id	mapping_justification
+HP:0009124	skos:exactMatch	MP:0000003	semapv:ManualMappingCuration
+HP:0000411	skos:exactMatch	sssom:NoTermFound	semapv:ManualMappingCuration
+```
+
+The second mapping means: `HP:0000411` does not have a corresponding semantic entity in the `object_source` that can be mapped using the `skos:exactMatch` property.
+
+- `sssom:NoTermFound` SHOULD NOT be used in any slot other than `subject_id` and `object_id`.
+  The use of `sssom:NoTermFound` in any slot other than `subject_id` and `object_id` is undefined.
+- When using `sssom:NoTermFound` in the `subject_id` slot, the `subject_source` slot SHOULD be defined.
+- When using `sssom:NoTermFound` in the `object_id` slot, the `object_source` slot SHOULD be defined.
--- a/examples/schema/no_term_found.sssom.tsv
+++ b/examples/schema/no_term_found.sssom.tsv
@@ -1,0 +1,16 @@
+#curie_map:
+#  HP: http://purl.obolibrary.org/obo/HP_
+#  MP: http://purl.obolibrary.org/obo/MP_
+#  obo: http://purl.obolibrary.org/obo/
+#  orcid: https://orcid.org/
+#mapping_set_id: https://w3id.org/sssom/commons/examples/no_term_found.sssom.tsv
+#creator_id:
+#  - orcid:0000-0002-7356-1779
+#subject_source: obo:hp
+#object_source: obo:mp
+#license: "https://creativecommons.org/publicdomain/zero/1.0/"
+#mapping_provider: "https://w3id.org/sssom/core_team"
+#comment: This is an example file for the SSSOM for illustration only. Its contents are entirely fabricated.
+subject_id	predicate_id	object_id	mapping_justification
+HP:0009124	skos:exactMatch	MP:0000003	semapv:ManualMappingCuration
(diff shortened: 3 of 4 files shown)


src/docs/spec-model.md
```diff
@@ -81,3 +81,41 @@ In addition, predicates from the following sources MAY also be encouraged:
 
 * any relation from the [Relation Ontology (RO)](https://obofoundry.org/ontology/ro.html);
 * any relation under [skos:mappingRelation](http://www.w3.org/2004/02/skos/core#mappingRelation) in the [Semantic Mapping Vocabulary](https://mapping-commons.github.io/semantic-mapping-vocabulary/).
+
+## Documenting unmapped elements
```
gouttegd, line 85: Sorry but I find this entire section way too confusing for something that is supposed to be normative, and it leaves some questions open.

My try:

## Representing unmapped entities

The special value `sssom:NoTermFound` MAY be used as the `object_id` of a mapping to explicitly state that the subject of said mapping cannot be mapped to any entity in the domain represented by the `object_source` slot.

Likewise, the `sssom:NoTermFound` value MAY be used as the `subject_id` of a mapping to state that the object of said mapping cannot be mapped to any entity in the domain represented by the `subject_source` slot.

When that special value is used as the `subject_id` (respectively `object_id`), the `subject_source` (respectively `object_source`) slot SHOULD be defined.

The `sssom:NoTermFound` value MUST NOT be used in any other slot than `subject_id` or `object_id`.

The meaning of the `NOT` predicate modifier in a mapping that refers to `sssom:NoTermFound` is unspecified.

When computing cardinality values (to fill the `mapping_cardinality` slot), mappings that refer to `sssom:NoTermFound` MUST be ignored.

src/docs/spec-model.md
```diff
@@ -81,3 +81,41 @@ In addition, predicates from the following sources MAY also be encouraged:
...
+
+The absence of a mapping in a mapping set can be interpreted in two ways:
+
+1. There is no mapping for that specific (absent) mapping that can be established
+2. The mapping was not established for some other reason, for example, the mapping set is a work in progress
+
+To circumvent this ambiguity, we can explicitly state that a specific term cannot be mapped 
+by using a SSSOM-built-in entity called `sssom:NoTermFound`.
+
+Example SSSOM TSV file:
```
gouttegd, line 95: Leave this kind of example for the user-facing documentation. They do not belong in the spec.

gouttegd (changes requested):
What you wrote is fine for the _documentation_, but not for the _spec_ IMO.

All the spec should do is warn implementers that they may come across the `sssom:NoTermFound` value, how they should interpret it, and what constraints apply to mappings where this value is used.

src/docs/spec-model.md
matentzn: Excellent. I am sorry that my work is a bit substandard compared to your rigour, I hope you are not annoyed you have to redo the proposals I make :) 

I have adopted your suggestion as is! Thank you!

gouttegd:
Two questions:

1) Do you want to explicitly forbid the case where _both_ the `subject_id` and the `object_id` are `sssom:NoTermFound`? Currently this would be perfectly allowed.

Such a mapping could be used to explicitly indicate that two domains (vocabularies, ontologies, databases, whatever) are completely disjoint and have _nothing_ in common. I suppose this would be a very infrequent need, but _a priori_ I see no reason to forbid that.

2) Do we need special rules for inverting mappings that refer to `sssom:NoTermFound`, when the mapping predicate has a distinct inverse predicate (e.g., the mapping predicate is `skos:narrowMatch`, which is inverted into `skos:broadMatch`)?

Consider the following mapping (justification omitted for brevity):

```
subject_id   subject_source   predicate_id       object_id           object_source
HP:1234      obo:hp           skos:narrowMatch   sssom:NoTermFound   obo:mp
```

Applying the classic inversion rules, this would be inverted into:

```
subject_id          subject_source   predicate_id       object_id   object_source
sssom:noTermFound   obo:mp           skos:broadMatch    HP:1234     obo:hp
```

I _think_ this is fine: the first mapping says “HP:1234 has no narrow match in MP”, the second says “nothing in MP has a broad match to HP:1234” – this seems correct to me, what do you think?

matentzn:
> Such a mapping could be used to explicitly indicate that two domains (vocabularies, ontologies, databases, whatever) are completely disjoint and have nothing in common. I suppose this would be a very infrequent need, but a priori I see no reason to forbid that.

I like the thinking. Do you think we should add this as a comment to the spec?

> Do we need special rules for inverting mappings that refer to sssom:NoTermFound, when the mapping predicate has a distinct inverse predicate (e.g., the mapping predicate is skos:narrowMatch, which is inverted into skos:broadMatch)?

Thank you for thinking of this! The example makes me was hard to think through mentally, but I think you are right, there is no problem. If HP:123 does not have a narrow match in MP, No term in MP can be a broad match to it.

gouttegd:
> I like the thinking. Do you think we should add this as a comment to the spec?

To the user-facing documentation, maybe. To the spec, no need, there’s nothing that implementations should be aware of.

gouttegd (dismissed):

hrshdhgd (changes requested):
Run `make all` same as https://github.com/mapping-commons/sssom/pull/386#pullrequestreview-2217867625

hrshdhgd (dismissed):

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 ## Next
 
 - Add the concept of "propagatable slots".
+- Add the entity reference `sssom:NoTermFound` of express the concept of an "unmapped entity" ([issue](https://github.com/mapping-commons/sssom/issues/28))
```
ehartley, line 6:
```suggestion
- Add the entity reference `sssom:NoTermFound` to express the concept of an "unmapped entity" ([issue](https://github.com/mapping-commons/sssom/issues/28))
```

src/docs/spec-model.md
```diff
@@ -81,3 +81,17 @@ In addition, predicates from the following sources MAY also be encouraged:
...
 * any relation from the [Relation Ontology (RO)](https://obofoundry.org/ontology/ro.html);
 * any relation under [skos:mappingRelation](http://www.w3.org/2004/02/skos/core#mappingRelation) in the [Semantic Mapping Vocabulary](https://mapping-commons.github.io/semantic-mapping-vocabulary/).
+
+## Representing unmapped entities
+
+The special value `sssom:NoTermFound` MAY be used as the `object_id` of a mapping to explicitly state that the subject of said mapping cannot be mapped to any entity in the domain represented by the `object_source` slot.
+
+Likewise, the `sssom:NoTermFound` value MAY be used as the `subject_id` of a mapping to state that the object of said mapping cannot be mapped to any entity in the domain represented by the `subject_source` slot.
+
+When that special value is used as the `subject_id` (respectively `object_id`), the `subject_source` (respectively object_source) slot SHOULD be defined.
```
ehartley, line 91:
```suggestion
When that special value is used as the `subject_id` (respectively `object_id`), the `subject_source` (respectively object_source) slot SHOULD be defined.
```
```suggestion
When that special value is used as the `subject_id` (respectively `object_id`), the `subject_source` (respectively `object_source`) slot SHOULD be defined.
```

src/docs/spec-model.md
```diff
@@ -81,3 +81,17 @@ In addition, predicates from the following sources MAY also be encouraged:
...
+
+The `sssom:NoTermFound` value MUST NOT be used in any other slot than subject_id or object_id.
```
ehartley, line 93:
```suggestion
The `sssom:NoTermFound` value MUST NOT be used in any other slot than `subject_id` or `object_id`.
```

src/docs/spec-model.md
```diff
@@ -81,3 +81,17 @@ In addition, predicates from the following sources MAY also be encouraged:
...
+
+The meaning of the NOT predicate modifier in a mapping that refers to `sssom:NoTermFound` is unspecified.
+
+When computing cardinality values (to fill the mapping_cardinality slot), mappings that refer to `sssom:NoTermFound` MUST be ignored.
```
ehartley, line 97:
```suggestion
When computing cardinality values (to fill the `mapping_cardinality` slot), mappings that refer to `sssom:NoTermFound` MUST be ignored.
```

PR #384, opened 2024-08-04
Add new "literal profile" solution based on "subject_label" and "object_label"

### LINKED ISSUE #234: SSSOM: Separate profile for Literal Mappings

matentzn:
Resolves #234

- [ ] Change base branch to `main`

This PR introduces an alternative model to the "literal mapping" proposal we have previously added. It is built on the following principles:

- Introducing a profile causes a needless overhead which requires specialised tooling
- We can achieve the same functionality with a _very_ minimal intervention in the main metadatamodel

I do understand that there are various opposing views on the need of a "literal" profile, but I think this super minimal intervention will satisfy both sides.

Huge thanks to @gouttegd who managed to steer this massive carrier ship after it has left the harbour. This is rarely successful and I am very happy we managed to make it!


src/sssom_schema/schema/sssom_schema.yaml
```diff
@@ -657,11 +660,31 @@ classes:
     - object_preprocessing
     - semantic_similarity_score
     - semantic_similarity_measure
+    - similarity_score
+    - similarity_measure
```
gouttegd, line 664: How do these two slots differ from the `semantic_`-prefixed ones?

src/sssom_schema/schema/sssom_schema.yaml
```diff
@@ -657,11 +660,31 @@ classes:
...
     - see_also
     - issue_tracker_item
     - other
     - comment
     class_uri: owl:Axiom
+    rules:
```
gouttegd, line 670: This only addresses the constraint on `subject_id / subject_label`. If we want to allow literals on the object side as well, we need another set of rules for `object_id / object_label`.

matentzn: Yes, I wanted some confirmation from some LinkML person (reached out on their slack) to see if this makes sense this way.

src/sssom_schema/schema/sssom_schema.yaml
matentzn: They are generailisations. Examples of something that is a `similarity_score` but not `semantic` is:

- Levenshtein Distance (Edit Distance): This measures the minimum number of single-character edits (insertions, deletions, or substitutions) required to change one string into another.
- Jaccard Similarity (on strings): This compares the overlap of characters or words between two strings, calculated as the size of the intersection divided by the size of the union of the two sets.
- Hamming Distance: This measures the number of positions at which corresponding characters in two strings of equal length are different.
- Jaro-Winkler Distance: This string metric measures edit distance while giving more favorable ratings to strings that match from the beginning.
- Longest Common Substring: This finds the longest string that is a substring of two or more strings, which can be used as a basis for similarity.

All of these are very typical in ontology / entity matching.

gouttegd: Sorry but this seems like another needless complication to me.

The two slots are defined in _exactly the same way_:

```yaml
similarity_score:
  description: A score between 0 and 1 to denote the similarity, where
    1 denotes equivalence.
  range: double
  minimum_value: 0.0
  maximum_value: 1.0
semantic_similarity_score:
  description: A score between 0 and 1 to denote the semantic similarity, where
    1 denotes equivalence.
  range: double
  minimum_value: 0.0
  maximum_value: 1.0
```

The corresponding `similarity_measure` and `semantic_similarity_measure` slots even use the same Jaccard similarity example:

```yaml
semantic_similarity_measure:
  description: The measure used for computing the the semantic similarity score.
    To make processing this field as unambiguous as possible, we recommend using 
    wikidata identifiers, but wikipedia pages could also be acceptable.
  range: string
  examples:
    - value: https://www.wikidata.org/wiki/Q865360
      description: (the Wikidata identifier for the Jaccard index measure).
  similarity_score:
    description: A score between 0 and 1 to denote the similarity, where
      1 denotes equivalence.
    range: double
    minimum_value: 0.0
    maximum_value: 1.0
similarity_measure:
  description: The measure used for computing the the similarity score.
    To make processing this field as unambiguous as possible, we recommend using 
    wikidata identifiers, but wikipedia pages could also be acceptable.
  range: string
  examples:
    - value: https://www.wikidata.org/wiki/Q865360
      description: (the Wikidata identifier for the Jaccard index measure).
```

So, again, why the two different slots? When you defined `semantic_similarity_{measure|score}` you seemingly didn’t have a problem with using it to report a Jaccard similarity, so why do we now need a distinct (non-semantic) `similarity_{measure|score}`?

matentzn: Because there are kinds of jaccard similarity which are semantic similarity measures, like for example "subclass-jaccard" (the number of overlapping parents).

I am happy though for this last push to deprecate the "semantic_*" variants and replace them by the generalised version in a new PR.

gouttegd: > I am happy though for this last push to deprecate the "semantic_*" variants and replace them by the generalised version in a new PR.

`semantic_similarity_{score|measure}` have been in the spec for quite some time, so if you do that (I have no objection, to be clear) I would recommend to add a new subsection to the [Compatibility with pre-1.0 versions](https://mapping-commons.github.io/sssom/spec-formats-tsv/#compatibility-with-pre-10-versions) section:

> ## semantic_similarity_score and semantic_similarity_measure
>
> Initial versions of this specification defined a `semantic_similarity_score` slot to store the semantic similarity, and a `semantic_similarity_measure` slot to describe how the the semantic similarity is assessed. In SSSOM 1.0, those slots were replaced by more generic `similarity_score` and `similarity_measure` slots.
>
> Upon encountering a `semantic_similarity_score` (respectively `semantic_similarity_measure`) slot, implementations supporting pre-1.0 versions MUST silently transform it into a `similarity_score` (respectively `similarity_measure`) slot. No changes on the value of the slot are required.

src/sssom_schema/schema/sssom_schema.yaml
matentzn: I asked for a full review from @cmungall so I committed the object rules as well

src/sssom_schema/schema/sssom_schema.yaml
matentzn: #385

gouttegd:
Two more questions:

1) Do we want to represent unmappable literal mappings? If yes, how?

This is basically a question of how the “literal profile” and `sssom:NoTermFound` should play together.

There are two distinct cases.

1.a) One side is a literal, the other side is `sssom:NoTermFound`.

Unless I am missing something, this case should be straightforward and should not require any special treatment. That’s a good thing, because it’s the case that is most likely to be useful: if you have to map a list of literals (say, a list of scRNAseq cluster names) to a target vocabulary (say, the CL ontology), you want to be able to say that a given literal could not be mapped to any term in the target vocabulary.

1.b) The literal and the “term not found” are on the same side.

Say you want to map a vocabulary (again, let’s say CL) to a list of literals. You want to say that a given term in CL is _not_ represented in the list of literals. We have a problem here, because the way to represent a “term not found” is to use `sssom:NoTermFound` as the `object_id` (assuming the literals are on the object side – it would not change anything if they were on the subject side), but `object_id` is _not_ used for a literal mapping (literals have no IDs, almost by definition).

So what to do here? I can imagine 3 possibilities:

1.b.1) We don’t support that case. Seems quite reasonable to me _a priori_, as I don’t think it is a very relevant use case (contrary to the 1.a case). That would be the current situation (with the profile as defined in this PR).

1.b.2) Use `sssom:NoTermFound` as the `object_id`, as we would do for non-literal mappings. This would mean that, even for literal mappings (where we are normally concerned only with `object_label`), we would need to lookup `object_id` just to check whether it contains the special `sssom:NoTermFound` value. It leaves open the question of what to put in the `object_label` slot in that case (we _have_ to put something in that slot: it cannot be empty).

1.b.3) Put in the `object_label` slot a special value similar to `sssom:NoTermFound`, intended to indicate that there is no mapped literal. It could even be the value `sssom:NoTermFound` itself (probably the easiest way).

2) Computing `mapping_cardinality` for literal mappings

Not really a question, but when computing the `mapping_cardinality`, if we are dealing with literal mappings, we must deal with the `_label` slots rather than the `_id` slots.

That is, for example: if we are dealing with a set of mappings with literals on the subject side and entities on the object side, then a cardinality of `1:n` means that one `subject_label` (NOT one `subject_id`) maps to several `object_id`.

That is, I think, quite obvious, but it should still be mentioned explicitly somewhere.

matentzn:
Ok, I will fold that in (think about it first). As an aside, I like it if actionable comments to be added as comments on the diffs so that they show up as "resolveable", so that they also block PRs being merged unless they are explicitly resolved. If there is no good place for such a comment you can just put in on the Changelog diff or on any other file that makes a little bit of sense.

gouttegd:
> As an aside, I like it if actionable comments to be added as comments on the diffs so that they show up as "resolveable"

Duly noted. :)

src/sssom_schema/schema/sssom_schema.yaml
matentzn, line 1: > Do we want to represent unmappable literal mappings? If yes, how?

I think so, yes. 

> 1.a) One side is a literal, the other side is sssom:NoTermFound.

> Unless I am missing something, this case should be straightforward and should not require any special treatment. 

Yes, I agree. No need to document anything else, perhaps, one day, in user facing docs!

> 1.b) The literal and the “term not found” are on the same side.
> So what to do here? I can imagine 3 possibilities:
> 1.b.1) We don’t support that case. Seems quite reasonable to me a priori, as I don’t think it is a very relevant use case (contrary to the 1.a case). That would be the current situation (with the profile as defined in this PR).

This is ok for me as an option, as I think the use case is so fringe that I cant imagine it arising right now. I wonder where this needs to be recorded?

> 1.b.2) Use sssom:NoTermFound as the object_id, as we would do for non-literal mappings. This would mean that, even for literal mappings (where we are normally concerned only with object_label), we would need to lookup object_id just to check whether it contains the special sssom:NoTermFound value. It leaves open the question of what to put in the object_label slot in that case (we have to put something in that slot: it cannot be empty).

That sounds too complex to me if I am honest.

> 1.b.3) Put in the object_label slot a special value similar to sssom:NoTermFound, intended to indicate that there is no mapped literal. It could even be the value sssom:NoTermFound itself (probably the easiest way).

I would prefer this, if we did this, where would we document that?

> Computing mapping_cardinality for literal mappings
> Not really a question, but when computing the mapping_cardinality, if we are dealing with literal mappings, we must deal with the _label slots rather than the _id slots.
> That is, for example: if we are dealing with a set of mappings with literals on the subject side and entities on the object side, then a cardinality of 1:n means that one subject_label (NOT one subject_id) maps to several object_id.
> That is, I think, quite obvious, but it should still be mentioned explicitly somewhere.

To avoid us circling over my mediocre spec writing ability, would you be able to give this a go? Or if not, tell me where it should go.

gouttegd: > > 1.b.1) We don’t support that case.
>
> […] I think the use case is so fringe that I cant imagine it arising right now.

Same here.

> > 1.b.2) Use `sssom:NoTermFound` as the `object_id`.
>
> That sounds too complex to me if I am honest.

Not necessarily _complex_, but cumbersome at the very least. Not a fan of that solution either.

> > 1.b.3) Put in the object_label slot a special value similar to sssom:NoTermFound.
>
> I would prefer this, if we did this, where would we document that?

If we do this, this would not require any big change, so how about:

* We make SSSOM 1.0 with the literal profile, but without support for the “1.b” case (i.e., in SSSOM 1.0 you can say that a literal cannot be mapped to an entity, but you cannot say that an entity cannot be mapped to a literal).
* _If_, after 1.0 is out, people start requesting that feature, we’ll reconsider supporting the “1.b“ case in SSSOM 1.x, by defining a special value for the `_label` slot.

?

gouttegd: About the mapping cardinality thing:

> would you be able to give this a go?

Will do. :) (which doesn’t mean I agree about your spec writing ability being “mediocre”)

> Or if not, tell me where it should go.

Since the “literal profile”, despite all the discussion we are having about it right now, remains a “minor” use case for SSSOM (which is still primarily intended for mapping between _entities_), I think the easiest way is to put everything related to the “literal profile” in a section of its own, and not refer to it at all everywhere else. The rest of the spec can be written as if the literal mappings did not exist.

So, for example, we can keep the description of `mapping_cardinality` in the model (“A string indicating whether this mapping is from a 1:1 (the subject_id maps to a single object_id), 1:n (the subject maps to more than one object_id), n:1, 1:0, 0:1 or n:n group.”) as it is. Then, in the section about literal mappings, we can explain what should be different in the case of literal mappings.

Seems much easier (both to write, and to read) than dispersing various bits of information about literal mappings everywhere in the spec.

src/sssom_schema/schema/sssom_schema.yaml
```diff
@@ -49,6 +49,11 @@ enums:
         meaning: rdfs:Datatype
       rdf property:
         meaning: rdf:Property
+      sssom literal:
```
matentzn, line 52: from @cmungall: Using spaces in enums might lead to funny encoded spaces in RDF serialisations.

gouttegd: All values on the `EntityType` enum (not only `sssom literal`) have had spaces since the beginning... Mildly opposed to changing that now.

gouttegd, line 52: The addition of `sssom literal` might require an explanation of what `rdfs literal` is intended for, and how those two differ.

matentzn: You are totally right.. I am wondering the same.. I dont think there is a point in rdfs literal, that was just some mistake I made in the early days of the schema. Should we:

1. Have "rdfs literal" do the job we assigned to "sssom literal" or
2. Remove "rdfs literal" to avoid confusion?

I am on the fence and cant decide. "sssom literal" sort of implies that it has a  "special place" in SSSOM. "rdfs literal is a bit ideosyncractic to the SemWeb world. But it is known, and "rdfs literal" is, in fact, expressing the same what "sssom literal is...

gouttegd: I am mildly inclined to choose (1), i.e., forget about `sssom literal` and use the already existing `rdfs literal`. But I’d be fine with (2) as well.

gouttegd, line 52: I just realized that the `predicate_type` slot is also using the `entity_type` enum. The spec for `sssom literal` (or `rdfs literal` if we chose to reuse that value) must make clear that it can only be used for the `subject_type` and `object_type` slot. We don’t want to allow the use of “literal predicates” (what would a literal predicate even be?).

matentzn: predicate_type needs its own enum.. Grrrrr. You are of course right.. Again

gouttegd: Not sure that a new enum for `predicate_type` would be necessary. If we have one enum for the subject and object entities and one enum for the predicate, there would be shared values between the two: `owl object property`, `owl data property`, `owl annotation property`, and `rdf property` were added to the `entity_type` enum primarily, I suppose, to be used to indicate the type of the predicate (in `predicate_type`), but they can also legitimately be used to indicate the type of the subject/object, if you are mapping properties (we do that for example in Uberon when we use a SSSOM set to replace some old properties by their modern RO equivalents).

So instead of having two enums with redundant values, we could simply say that only a subset of the `entity_type` values (the 4 listed above) can be used in the `predicate_type` slot. `owl class`, `owl named individual`, `skos concept`, etc. and `sssom literal` MUST NOT be used in `predicate_type`.

matentzn: @graybeal thanks for weighing in, this pushed me firmly over the edge of indecision towards using rdfs literal!

matentzn: Agreed.

matentzn: It opens another can of worms that the value, in a SSSOM TSV file, is going to be "rdfs literal" and in RDF, it is going to be `rdfs:Literal`. My bad for not defining `*_type` slots as linkml enums rather than entity references.. That seems to far gone now to change. At least, the model clearly binds the two together (using the `meaning` property).

matentzn: In any case, the issue here is fixed now, if you are satisfied @gouttegd please resolve conversation.

gouttegd: > It opens another can of worms that the value, in a SSSOM TSV file, is going to be "rdfs literal" and in RDF, it is going to be rdfs:Literal.

I don’t see how that is a problem. The same value will be serialised differently in different serialisation formats. So what?

matentzn: No need to restart that debate now.. I let a formalism (LinkML) determine the serialisation, not right. There are good reasons to use rdfs:Literal instead of "rdfs literal", like the fact that it is instantly recognisable as to what it means.

But that ship has sailed, and yes, there are good reasons to use "human readable" enums here in the *_type fields as well! So, its not too bad.

gouttegd: I am not debating anything, I just don’t see what makes you think that this is a “can of worms” that the same value can have different serialisations. For me this is a non-issue.

matentzn: Yes, didnt mean to say, "no need for you Damien to debate with me", I meant to say "no need to resurface this past decision and debate it with my younger self :P Non-issue indeed!

src/sssom_schema/schema/sssom_schema.yaml
```diff
@@ -45,10 +45,15 @@ enums:
         meaning: rdfs:Class
       rdfs literal:
         meaning: rdfs:Literal
+        description: "'rdfs literal' used in SSSOM files to express that an entity is represented by a literal."
```
matentzn, line 48: Better description needed.

src/sssom_schema/schema/sssom_schema.yaml
gouttegd: Done in #387 (to be merged to _this_ PR if you agree with it).

src/sssom_schema/schema/sssom_schema.yaml
gouttegd: I’ve put a tentative one in #387.

src/sssom_schema/schema/sssom_schema.yaml
gouttegd: This is solved as far as `rdfs literal` is concerned (the in-schema documentation explicitly states that `rdfs literal` cannot be used in `predicate_type`).

However we want to solve the issue for the other values in `entity_type` (including, possibly, by creating a new distinct enum specifically for `predicate_type`), this has no longer anything to do with literal mappings and would belong to a new, dedicated ticket.

gouttegd:
The SSSOM-Py test suite fails on three tests:

* `test_utils.py::TestUtils::test_get_dict_from_mapping` fails because it expects a `semantic_similarity_score` slot. This has nothing to do with this PR and is instead obviously the result of #386, which was _not_ tested against SSSOM-Py before being merged.

* `test_parsers.py::TestParse::test_parse_alignment_xml` fails when parsing this:

```xml
<Cell>"
  <entity1 rdf:resource="http://randomurlwithnochancetobeinprefixmap.org/ID_123"/>
  <entity2 rdf:resource="http://purl.obolibrary.org/obo/WBbt_0005815"/>
  <measure rdf:datatype="xsd:float">0.5</measure>
  <relation>=</relation>
</Cell>
```

My understanding is that, prior to this PR, after some checking against a prefix map, the `entity1` ends up being being empty/None/null, and prior to that PR SSSOM-Py could count on the `Mapping` constructor to automatically reject the mapping as invalid (since the Python generated code included a non-null check because of the `required=true` attribute on `subject_id`). With this PR, `subject_id` is no longer required, so the mapping is “successfully” constructed – which is wrong, and the test catches that.

It looks like SSSOM-Py cannot simply rely on the LinkML runtime implementing the `rules` set forth in this PR, so it will have to implement its own validation of `subject_id` / `object_id`.

* `test_parsers.py::TestParse::test_parse_obographs`: I have not looked in details, but it’s most likely a similar problem as above.

gouttegd:
The issues above are putatively fixed by https://github.com/mapping-commons/sssom-py/pull/552.

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 - Add the concept of "propagatable slots".
 - Add the `curie_map` to the model (instead of it being a specificity of the SSSOM/TSV format).
 - Add the concept of "extension slots".
+- Add the concept of "literal mappings".
```
matentzn, line 8: - Replace the previous idea of a "literal profile" that is separate from the main SSSOM standard with a simple flag that allows publishing literal mappings in the standard SSSOM format

src/docs/spec-model.md
```diff
@@ -92,6 +92,25 @@ In addition, predicates from the following sources MAY also be encouraged:
 * any relation under [skos:mappingRelation](http://www.w3.org/2004/02/skos/core#mappingRelation) in the [Semantic Mapping Vocabulary](https://mapping-commons.github.io/semantic-mapping-vocabulary/).
 
 
+## Literal mappings
+
+<a id="literal-mappings"></a>
+
+The SSSOM model is primarily intended to represent mappings between semantic entities. However, it may also be used to represent mappings where one side is a literal string that does not have an identifier of its own. Any such mapping is henceforth called a _literal mapping_.
```
matentzn, line 99: "where one side" - If we want to specifically prohibit literal - literal mappings we should probably make a very clear in the spec, and add this to the schema as a separate postcondition (if subject_type literal --> subject_label required AND object_if required)?

src/sssom_schema/schema/sssom_schema.yaml
```diff
@@ -45,10 +45,16 @@ enums:
         meaning: rdfs:Class
       rdfs literal:
         meaning: rdfs:Literal
+        description: This value indicate that the entity being mapped is not a semantic entity with a distinct identifier, but is instead represented entirely by its literal label. This value MUST NOT be used in the predicate_type slot.
```
matentzn, line 48: Its very easy to add the predicate_type restriction to the model directly (using `none_of`). Shall I do this for completeness or is the spec comment enough? (Expecting that tool implementors have to add that manually)

matentzn (commented):
Sorry @gouttegd , I was out these past days, now I am getting back :P Some more comments, looking at sssom py now!

CHANGELOG.md
gouttegd: Since the “previous idea of a ‘literal profile’” had never been officially announced in the changelog, and that profile had never been usable anyway (having never been connected to the rest of the spec), I am not sure we need to mention it. No objection, though.

src/docs/spec-model.md
gouttegd: Do we want to forbid literal-to-literal mappings? I have no objection to that, but I see no reason to do it, either.

If we do, I wouldn’t bother expressing that as a LinkML postcondition – they are useless as far as running code is concerned.

src/sssom_schema/schema/sssom_schema.yaml
gouttegd: I’d rather do that as a separate PR in which we also do that for the other `entity_type` values that should not be used in `predicate_type`, but no objection.

src/docs/spec-model.md
matentzn: I would personally not forbid it, for no other reasons other than that there is no strong _technical_ reason to do so. Literal to literal mappings may sound useless now, but I can see some use cases for them like data model mappings (where enums are represented as strings). So my inclination would be to simply change the phrasing `where one side is a literal string` to `where at least one side is a literal string`.

src/sssom_schema/schema/sssom_schema.yaml
matentzn: I agree, lets not scope creep this PR! If we want to do this formally, we do it in the next iteration. We have already documented it, so its not a "schema change" when we then do it.

CHANGELOG.md
matentzn: Ok! Your judgment. I am fine either way.

src/docs/spec-model.md
gouttegd: > So my inclination would be to simply change the phrasing where one side is a literal string to where at least one side is a literal string.

OK, will do.

src/docs/spec-model.md
```diff
@@ -92,6 +92,25 @@ In addition, predicates from the following sources MAY also be encouraged:
...
+
+<a id="literal-mappings"></a>
+
+The SSSOM model is primarily intended to represent mappings between semantic entities. However, it may also be used to represent mappings where at least one side is a literal string that does not have an identifier of its own. Any such mapping is henceforth called a _literal mapping_.
+
+To represent a mapping whose subject (resp. object) is a literal:
+
+* the `subject_type` (resp. `object_type`) slot is set to `rdfs literal`;
+* the `subject_label` (resp. `object_label`) slot is set to the literal itself;
+* the `subject_id` (resp. `object_id`) slot is left empty.
```
matentzn, line 105: Because this list is not using `MUST` and `SHOULD` phrases, it is left unclear if you can switch to literal mode and still have something in the id slot. Thinking of this a bit, I actually would, counterintuitively, favour this slightly, because it would give me a natural way to encode enum mappings (no need to be normative here). I also do not want to, right this moment, think it through if this is a bad idea or not (it might be).

subject_type: "rdfs literal"
subject_id: http://w3id.org/example/clinicaldatamodel/gender
subject_label: "FEMALE"
object_id: `PATO:0000383`

Again, I am not saying this is a good idea, but maybe its worth keeping the door open?

The alternative would be to have a seperate slot, `subject_label_context` or some such, in a future version.

In any case, what is actionable:

Either we should clearly say that it `MUST` be empty, or clearly say `SHOULD` be empty.

gouttegd: > still have something in the id slot. Thinking of this a bit, I actually would, counterintuitively, favour this slightly

So do I. For at least two reasons:

* Symmetry: For a non-literal mapping (where what matters is the `subject/object_id`), it is of course still perfectly valid to have a `subject/object_label`, so I feel like it should also be allowed, for a literal mapping (where what matters is the `subject/object_label`), to have something in `subject/object_id`.
* Somehow, even a literal could have _some kind_ of identifier, which may not be a “proper” semantic identifier (for example, it may be only meaningful for the person who compiled the list of literals) but that you may still want to record.

For example, if I map a list of scRNAseq cluster names to CL cell types, I _may_ want to do something like this:

```
#subject_type: rdfs literal
#object_source: obo:CL
subject_id   subject_label   predicate_id      object_id
cluster:1    astrocyte       skos:exactMatch   CL:1234567
cluster:2    oligodendrocyte skos:exactMatch   CL:2345678
...
```

These are “literal mappings” because the clusters are not really entities in any database or ontology, they are just entries in a local list that is only meaningful to me (at least until the dataset is published), and what is important for the matching here is really the name I gave to those clusters.

Allowing this kind of “pseudo-IDs” would also make it easier to convert those mappings to proper entity-to-entity mappings if and once the clusters get somehow published later and get assigned proper permanent IDs.

> Either we should clearly say that it MUST be empty, or clearly say SHOULD be empty.

I’d rather say “MAY be left empty”. No reason to strongly recommend that it should be left empty. The point is that the `subject/object_id` slot is basically _irrelevant_ for literal mappings, so do whatever you want with it (fill it or leave it empty, it doesn’t matter).

gouttegd: About “encoding enum mapping”:

> ```
> subject_type: "rdfs literal"
> subject_id: http://w3id.org/example/clinicaldatamodel/gender
> subject_label: "FEMALE"
> object_id: PATO:0000383
> ```

You mean using `subject_id` to identify the enum whose values you are mapping? Wouldn’t `subject_source` be more appropriate for that?

matentzn (commented):
APPROVED (I cant formally approve it, so @gouttegd you will have to).

src/docs/spec-model.md
matentzn: Yes, probably!

gouttegd (approved):

jamesamcl (approved):

PR #386, opened 2024-08-04
Similarity score

### LINKED ISSUE #385: Deprecate semantic_similarity_* slots in favour of the more generic similarity_* slots
`similarity_measure` and `similarity_score` both already exist in the model from a different line of work. 

We frequently want to recording similarity scores for measures that are not semantic.

Examples of something that is a `similarity_score` but not `semantic` are:

- Levenshtein Distance (Edit Distance): This measures the minimum number of single-character edits (insertions, deletions, or substitutions) required to change one string into another.
- Jaccard Similarity (on strings): This compares the overlap of characters or words between two strings, calculated as the size of the intersection divided by the size of the union of the two sets.
- Hamming Distance: This measures the number of positions at which corresponding characters in two strings of equal length are different.
- Jaro-Winkler Distance: This string metric measures edit distance while giving more favorable ratings to strings that match from the beginning.
- Longest Common Substring: This finds the longest string that is a substring of two or more strings, which can be used as a basis for similarity.

All of these are very typical in ontology / entity matching.

I move to replacing the semantic_similarity_* variants entirely with the more generic ones.

matentzn:
Fixes #385

### DIFF
--- a/src/docs/spec-formats-tsv.md
+++ b/src/docs/spec-formats-tsv.md
@@ -179,6 +179,11 @@ Any other value in the `match_term_type` slot MUST be treated as an error.
 
 If the set already contains `subject_type` and `object_type` slots, any `match_term_type` slot can be silently ignored.
 
+#### semantic_similarity_score and semantic_similarity_measure
+
+Initial versions of this specification defined a `semantic_similarity_score` slot to store the semantic similarity, and a `semantic_similarity_measure` slot to describe how the the semantic similarity is assessed. In SSSOM 1.0, those slots were replaced by more generic `similarity_score` and `similarity_measure` slots.
+
+Upon encountering a `semantic_similarity_score` (respectively semantic_similarity_measure) slot, implementations supporting pre-1.0 versions MUST silently transform it into a `similarity_score` (respectively `similarity_measure`) slot. No changes on the value of the slot are required.
 
 ## Canonical SSSOM/TSV format
 
--- a/src/sssom_schema/schema/sssom_schema.yaml
+++ b/src/sssom_schema/schema/sssom_schema.yaml
@@ -556,20 +556,6 @@ slots:
       - https://github.com/mapping-commons/sssom/issues/166
       - https://github.com/mapping-commons/sssom/pull/258
       - https://github.com/mapping-commons/sssom/blob/master/examples/schema/curation_rule_text.sssom.tsv
-  semantic_similarity_score:
-    description: A score between 0 and 1 to denote the semantic similarity, where
-      1 denotes equivalence.
-    range: double
-    minimum_value: 0.0
-    maximum_value: 1.0
-  semantic_similarity_measure:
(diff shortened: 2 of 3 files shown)


gouttegd (approved):
Looks all good to me.

hrshdhgd (changes requested):
You may want to update files that are dependent on this YAML change. For example [sssom_schema.py](https://github.com/mapping-commons/sssom/blob/master/src/sssom_schema/datamodel/sssom_schema.py) and others (if any). Run `make all` I think?

hrshdhgd (dismissed):

src/docs/spec-formats-tsv.md
```diff
@@ -179,6 +179,11 @@ Any other value in the `match_term_type` slot MUST be treated as an error.
 
 If the set already contains `subject_type` and `object_type` slots, any `match_term_type` slot can be silently ignored.
 
+#### semantic_similarity_score and semantic_similarity_measure
+
+Initial versions of this specification defined a `semantic_similarity_score` slot to store the semantic similarity, and a `semantic_similarity_measure` slot to describe how the the semantic similarity is assessed. In SSSOM 1.0, those slots were replaced by more generic `similarity_score` and `similarity_measure` slots.
+
+Upon encountering a `semantic_similarity_score` (respectively semantic_similarity_measure) slot, implementations supporting pre-1.0 versions MUST silently transform it into a `similarity_score` (respectively `similarity_measure`) slot. No changes on the value of the slot are required.
```
ehartley, line 186:
```suggestion
Upon encountering a `semantic_similarity_score` (respectively `semantic_similarity_measure`) slot, implementations supporting pre-1.0 versions MUST silently transform it into a `similarity_score` (respectively `similarity_measure`) slot. No changes on the value of the slot are required.
```

src/sssom_schema/schema/sssom_schema.yaml
```diff
@@ -556,34 +556,36 @@ slots:
...
     minimum_value: 0.0
     maximum_value: 1.0
+    see_also:
+      - https://github.com/mapping-commons/sssom/issues/385
+      - https://github.com/mapping-commons/sssom/pull/386
+      - https://github.com/mapping-commons/sssom/blob/master/examples/schema/similarity_score.sssom.tsv
   similarity_measure:
-    description: The measure used for computing the the similarity score.
+    description: The measure used for computing a similarity score.
+      The score is meant to be used in conjunction with the similarity_score field, to document,
```
ehartley, line 573:
```suggestion
      This field is meant to be used in conjunction with the similarity_score field, to document,
```

matentzn:
Thank you @ehartley @hrshdhgd @gouttegd for your reviews! I merged this, because @hrshdhgd and @gouttegd have approved, and @ehartley fixed two typos - no need for re-review for these (judgement call - let me know if we should not do that for procedural reasons).

Issue #189, opened 2022-06-03
Make SSSOM 1.0 release


matentzn:
All features are merged now.

Before we make a formal release we will have to deal with these two:
- [x] https://github.com/mapping-commons/semantic-mapping-vocabulary/issues/2
- [X] https://github.com/mapping-commons/sssom-py/issues/256 (to make sure no problems with toolkit from current standard), cc @hrshdhgd 
- [X] Update all documentation/specification/tutorial pages to reflect these changes DONE: https://github.com/mapping-commons/sssom/pull/205/files
- [x]  Update changelog https://github.com/mapping-commons/sssom/blob/master/CHANGELOG.md

Target date for 1.0 release: June 10.

matentzn:
I added now another intermediate release 0.9.4 which reflects all the changes we wanted to get in for SSSOM 1.0. 
https://github.com/mapping-commons/sssom/releases/tag/0.9.4

I still want to leave at least a window open for the review from @graybeal and @ernestojimenezruiz on the CV above, but I have now updates basically everything to use the vocab as is.

saubin78:
Hi! Do you plan to release the new version of SSSOM soon?

matentzn:
We are still rolling with the 0.9.X series, are you asking because of the curation_rules?

matentzn:
It is top of my list; I will make a new release at the end of next week, is that ok?

saubin78:
Great! Thanks Nico

matentzn:
After 4 years, the day has come.

With 79 contributors writing more than 1200 comments over the last 4 years, we have done it!

https://github.com/mapping-commons/sssom/releases/tag/v1.0.0

## Contribution stats for SSSOM from 2020-04-20 to 2024-08-09

Created with [grainyhead](https://github.com/gouttegd/grainyhead).

| Event | Total | Internal | Internal (%) | External | External (%) |
| -------- | -------- | -------- | -------- | -------- | -------- |
| Contributors | 79 | 21 |  26.58 | 58 |  73.42 |
| Issues opened | 218 | 172 |  78.90 | 46 |  21.10 |
| Issues closed | 114 | 110 |  96.49 | 4 |  3.51 |
| Pull requests opened | 154 | 145 |  94.16 | 9 |  5.84 |
| Pull requests closed | 152 | 146 |  96.05 | 6 |  3.95 |
| Pull requests merged | 129 | 126 |  97.67 | 3 |  2.33 |
[...]

PR #388, opened 2024-08-09
Update generated SSSOM schema files

matentzn:
Last commit before #189!

This is the usual update with `make all` (generating all downstream python files).

This is going to be the final model update before sssom 1.0.


project/graphql/sssom_schema.graphql
```diff
@@ -7,13 +8,13 @@ type ExtensionDefinition
 
 type Mapping
   {
-    subjectId: EntityReference!
+    subjectId: EntityReference
```
hrshdhgd, line 11: What did the `!` mean?

hrshdhgd (approved):

project/graphql/sssom_schema.graphql
matentzn: Required? I am assuming!

gouttegd: It meant the field was _non-nullable_ (see [GraphQL’s doc](https://graphql.org/learn/schema/#object-types-and-fields), which was indeed before #384 the case since the model defined that field as `required`.

gouttegd (approved):

project/graphql/sssom_schema.graphql
matentzn: Thanks!

PR #398, opened 2024-11-03
Use uv to make reproducible calls to poetry

### LINKED ISSUE #397: `poetry install` fails
I'm on an m3 mac running mac os 14. I installed poetry with `uv tool install poetry`.

```console
$ poetry --version
Poetry (version 1.8.4)
```

I got the following error while trying to install. This seems like an version of pydantic-core

```console
$ poetry install
Installing dependencies from lock file

Package operations: 75 installs, 0 updates, 0 removals

  - Installing pydantic-core (2.16.3): Failed

  ChefBuildError

  Backend subprocess exited when trying to invoke build_wheel
  
  Running `maturin pep517 build-wheel -i /private/var/folders/6k/r2gmmc213vl6h7bv20x624xw0000gn/T/tmp2_z1mb_t/.venv/bin/python --compatibility off`
  💥 maturin failed
    Caused by: Cargo metadata failed. Do you have cargo in your PATH?
    Caused by: No such file or directory (os error 2)
  Error: command ['maturin', 'pep517', 'build-wheel', '-i', '/private/var/folders/6k/r2gmmc213vl6h7bv20x624xw0000gn/T/tmp2_z1mb_t/.venv/bin/python', '--compatibility', 'off'] returned non-zero exit status 1
  

  at ~/Library/Application Support/uv/tools/poetry/lib/python3.13/site-packages/poetry/installation/chef.py:164 in _prepare
      160│ 
      161│                 error = ChefBuildError("\n\n".join(message_parts))
      162│ 
      163│             if error is not None:
    → 164│                 raise error from None
      165│ 
      166│             return path
      167│ 
      168│     def _prepare_sdist(self, archive: Path, destination: Path | None = None) -> Path:

Note: This error originates from the build backend, and is likely not a problem with poetry but with pydantic-core (2.16.3) not supporting PEP 517 builds. You can verify this by running 'pip wheel --no-cache-dir --use-pep517 "pydantic-core (==2.16.3)"'.

  - Installing rpds-py (0.18.0): Failed

  ChefBuildError

  Backend subprocess exited when trying to invoke build_wheel
  
  Running `maturin pep517 build-wheel -i /private/var/folders/6k/r2gmmc213vl6h7bv20x624xw0000gn/T/tmpz1opyj4b/.venv/bin/python --compatibility off`
  💥 maturin failed
    Caused by: Cargo metadata failed. Do you have cargo in your PATH?
    Caused by: No such file or directory (os error 2)
  Error: command ['maturin', 'pep517', 'build-wheel', '-i', '/private/var/folders/6k/r2gmmc213vl6h7bv20x624xw0000gn/T/tmpz1opyj4b/.venv/bin/python', '--compatibility', 'off'] returned non-zero exit status 1
  

  at ~/Library/Application Support/uv/tools/poetry/lib/python3.13/site-packages/poetry/installation/chef.py:164 in _prepare
      160│ 
      161│                 error = ChefBuildError("\n\n".join(message_parts))
      162│ 
      163│             if error is not None:
    → 164│                 raise error from None
      165│ 
      166│             return path
      167│ 
      168│     def _prepare_sdist(self, archive: Path, destination: Path | None = None) -> Path:

Note: This error originates from the build backend, and is likely not a problem with poetry but with rpds-py (0.18.0) not supporting PEP 517 builds. You can verify this by running 'pip wheel --no-cache-dir --use-pep517 "rpds-py (==0.18.0)"'.

```

cthoyt:
Closes #397

Because poetry appears to be highly dependent on the version of Python, this PR updates the makefile to reproducibly install poetry into a given version of python using uv.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,8 @@ SHELL := bash
 .SUFFIXES:
 .SECONDARY:
 
-RUN = poetry run
+POETRY = uv tool run --python 3.11 poetry
+RUN = $(POETRY) run
 # get values from about.yaml file
 SCHEMA_NAME = $(shell sh ./utils/get-value.sh name)
 SOURCE_SCHEMA_PATH = $(shell sh ./utils/get-value.sh source_schema_path)
@@ -38,7 +39,7 @@ status: check-config
 setup: install gen-project gendoc git-init-add
 
 install:
-	poetry install
+	$(POETRY) install
 .PHONY: install
 
 all: gen-project gendoc gen-excel get-context
@@ -74,7 +75,7 @@ examples/%.ttl: src/data/examples/%.yaml
 	$(RUN) linkml-convert -P EXAMPLE=http://example.org/ -s $(SOURCE_SCHEMA_PATH) -C Person $< -o $@
 
 upgrade:
-	poetry add -D linkml@latest
+	$(POETRY) add -D linkml@latest
 
 # Test documentation locally
 serve: mkd-serve


matentzn:
QC failures related to this PR?

gouttegd:
Why using yet another tool (`uv`) just to force Poetry to use a given version of Python

Poetry has the `env use` command for that:

```sh
poetry env use 3.11
poetry install
```

cthoyt:
Issue is poetry needs to be installed on 3.11.

gouttegd:
Wait, Poetry only runs correctly on Python 3.11? Seriously?

matentzn:
@cthoyt what do we want to do with this PR here? Any plans to wrap/finalise?

gouttegd:
FWIW, the idea of using _both_ UV and Poetry (using UV to install Poetry to manage the project) seems completely ridiculous to me.

Either we stick to Poetry or we switch completely to UV (I would have no objection either way), but mixing two project/package managers can only invite more trouble.

cthoyt:
This one was mostly to demonstrate that poetry is hard to work with. I think everyone at this point agrees

I would like to switch fully over to uv. Since I made some other PRs updating the metadata do be more portable and less poetry-specific, we could do that in a different PR.

cthoyt:
uv doesn't support dynamic versioning (and might not ever)

My personal taste is also to use explicit versions instead of trying to mix in the version control system, which to me is a weird mixture of separate concerns

this might just be an artifact of a choice made long ago in the oboverse that we could revisit

PR #403, opened 2024-12-02
Add 'composed entity' value to the EntityType enum.

### LINKED ISSUE #402: New EntityType value to identify composed entities
There has been quite a few discussions about whether it should be possible to represent mappings involving “complex” or “composed” or “post-coordinated” entities – where at least one side of the mapping is not a single entity but a composition of several entities — and if so, _how_ to represent such mappings.

For example:

* #36 
* #108
* #180 

In #108 in particular, the discussion seems to have stabilized around the general idea of _encoding the post-coordinated entity within the subject (or object) ID_, using a construct such as, for example, `obo.pattern:anatomical_trait?anatomical_entity=UBERON:123&disease=PATO:123` (or various variations of it). This has led to the proposal for [URI Expressions](https://github.com/monarch-initiative/uri-expression-language), which is the exact same idea but using a form of JSON encoding rather than URI parameters.

Regardless of _how_ the post-coordinated entities are ultimately encoded within a ID, the main benefit of this general approach is that it requires almost no change to SSSOM at all. Subject and object IDs in SSSOM are always opaque strings, so the fact that one ID happens to encode a post-coordinated entity using some elegant/ugly hack is completely irrelevant. SSSOM implementations can just pass the ID as it is, and leave it to applications built on top of them to eventually “unpack” the ID (stop treating it as opaque) and get access to the separate entities that made up the post-coordinated one.

*But* if this approach is used (in fact that approach can _already_ be used right now, precisely because it does not require any explicit support in SSSOM), SSSOM should at least allow to explicitly indicate that the subject or object of a mapping is a post-coordinated entity.

SSSOM allows to indicate the type of “entity being mapped” with the `subject_type` and `object_type` slots, which accept a value from the [EntityType](https://mapping-commons.github.io/sssom/EntityTypeEnum/) enumeration. Currently, there is no value in that enumeration that can be suitable to mark the type of a post-coordinated entity; it would be wrong to say that a post-coordinated entity is a “owl class”, or a “owl object property”, or a “skos concept”, or a “rdfs resource”, or any of the other available type. A post-coordinated entity is a _sui generis_ type of entity, which should have its own representation in the `EntityType` enumeration.

I therefore propose that a new value, tentatively named `composed entity`, be added to the `EntityType` enumeration.

gouttegd:
Resolves #402

- [ ] ~~`docs/` have been added/updated if necessary~~ (not needed; the new enum value is documented directly in the model’s `description` field)
- [ ] ~~tests have been added/updated (if applicable)~~ (nothing to test; the new value does not change anything to the expected behaviours)

This PR adds a new value `composed entity` to the `EntityType` enumeration to indicate, as proposed in #402, that an ID is intended to represent a composed (aka “complex”, aka “post-coordinated”) entity that involves several individual entities.

### DIFF
--- a/src/sssom_schema/schema/sssom_schema.yaml
+++ b/src/sssom_schema/schema/sssom_schema.yaml
@@ -45,7 +45,7 @@ enums:
         meaning: rdfs:Class
       rdfs literal:
         meaning: rdfs:Literal
-        description: This value indicate that the entity being mapped is not a semantic entity with a distinct identifier, but is instead represented entirely by its literal label. This value MUST NOT be used in the predicate_type slot.
+        description: This value indicates that the entity being mapped is not a semantic entity with a distinct identifier, but is instead represented entirely by its literal label. This value MUST NOT be used in the predicate_type slot.
         see_also:
         - https://mapping-commons.github.io/sssom/spec-model/#literal-mappings
         - https://github.com/mapping-commons/sssom/issues/234
@@ -54,6 +54,11 @@ enums:
         meaning: rdfs:Datatype
       rdf property:
         meaning: rdf:Property
+      composed entity:
+        description: This value indicates that the entity ID does not represent a single entity, but a composite involving several individual entities. This value MUST NOT be used in the predicate_type slot. This specifications does not prescribe how an ID representing a composite entity should be interpreted; this is left at the discretion of applications.
+        see_also:
+        - https://github.com/mapping-commons/sssom/issues/402
+        - https://github.com/mapping-commons/sssom/blob/master/examples/schema/composite-entities.sssom.tsv
         
   predicate_modifier_enum:
     permissible_values:
--- a/src/sssom_schema/schema/sssom_schema.yaml
+++ b/src/sssom_schema/schema/sssom_schema.yaml
@@ -54,7 +54,7 @@ enums:
         meaning: rdfs:Datatype
       rdf property:
         meaning: rdf:Property
-      composed entity:
+      composed entity expression:
         description: This value indicates that the entity ID does not represent a single entity, but a composite involving several individual entities. This value MUST NOT be used in the predicate_type slot. This specifications does not prescribe how an ID representing a composite entity should be interpreted; this is left at the discretion of applications.
         see_also:
         - https://github.com/mapping-commons/sssom/issues/402
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## Next
 
+- Add `composed entity` as a new value in the `EntityType` enumeration ([issue](https://github.com/mapping-commons/sssom/issues/402)).
 - TBD
 
 ## SSSOM version 1.0.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## Next
 
-- Add `composed entity` as a new value in the `EntityType` enumeration ([issue](https://github.com/mapping-commons/sssom/issues/402)).
+- Add `composed entity expression` as a new value in the `EntityType` enumeration ([issue](https://github.com/mapping-commons/sssom/issues/402)).
 - TBD
 
 ## SSSOM version 1.0.0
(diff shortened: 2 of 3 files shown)


gouttegd:
@matentzn Two things that might be worth discussing:

(A) I have not specified a `meaning` for the new value. First because it’s unclear to me what the… meaning of the `meaning` slot actually is ([LinkML’s documentation](https://linkml.io/linkml-model/latest/docs/meaning/) is not really helpful here, I believe), and in particular whether it is supposed to be normative or merely informative.

If a `meaning` is absolutely required, a possible candidate could be `owl:ClassExpression`, but I don’t like the fact that it seems to tie the concept of composed entity to OWL.

(B) I have explicitly forbidden the use of `composed entity` in the `predicate_type` slot. That is because, while it could theoretically be possible to use “post-coordinated predicates”, I don’t think this would be a good idea and I’d rather cut it short immediately. But I am open to counter-arguments on this.

matentzn (dismissed):
I absolutely love it! I will get maybe two more reviews on this.

src/sssom_schema/schema/sssom_schema.yaml
```diff
@@ -54,6 +54,11 @@ enums:
         meaning: rdfs:Datatype
       rdf property:
         meaning: rdf:Property
+      composed entity:
+        description: This value indicates that the entity ID does not represent a single entity, but a composite involving several individual entities. This value MUST NOT be used in the predicate_type slot. This specifications does not prescribe how an ID representing a composite entity should be interpreted; this is left at the discretion of applications.
```
cmungall, line 58: Do we prefer `composed entity` over something like `{composed,composite} entity expression`?

I wish there was an element in the OWL vocabulary we could map to. the OWLAPI has a notion of "AnonymousClassExpression" but that doesn't map to anything in the OWL spec (https://www.w3.org/TR/owl2-syntax/#Class_Expressions). And in any case, a composed entity need not be a class expression.

I think the OWLAPI's use of "anonymous" for compound expressions is quite unique here. The more typical mathematic adjective is "composed" or "compound" or "complex" AFAICT. But I think adding "expression" makes it clear this does not denote a named entity in the more typical sense?

gouttegd: > Do we prefer composed entity over something like {composed,composite} entity expression?

I don’t really have a strong preference here. My main concerns with the choice of the name was (1) to convey the idea that it is not a “simple” entity and (2) to avoid tying the concept to any particular representation or encoding (such as the “URI Expression” proposal mentioned in the #402 ticket). If people think that the first aim is better achieved by adding ` expression`, I have no objection to that.

matentzn (approved):
Let's go with this - I think we are the main stakeholders here anyways, and we have had this PR open one work week. i will merge on Tuesday and create a release.

gouttegd:
> If a meaning is absolutely required, a possible candidate could be owl:ClassExpression, but I don’t like the fact that it seems to tie the concept of composed entity to OWL.

I just realised that the `meaning` slot in enum values is used by SSSOM-Py when serialising to RDF. For example, a mapping with a `subject_type` set to `owl class` will be serialised in RDF as:

```
[ a owl:Axiom ;
    owl:annotatedProperty THE:PREDICATE ;
    owl:annotatedSource THE:SUBJECT ;
    owl:annotatedTarget THE:OBJECT ;
    sssom:subject_type owl:Class ]
```

With no `meaning` set for `composed entity expression`, that particular value would be serialised as a literal string instead:

```
[ a owl:Axiom ;
    owl:annotatedProperty THE:PREDICATE ;
    owl:annotatedSource THE:SUBJECT ;
    owl:annotatedTarget THE:OBJECT ;
    sssom:subject_type "composed entity expression" ]
```

sierra-moxon:
Do you think it would be worthwhile to constrain `subject_id` so that the composition has a standard format? 
`SCHEMA:0001/(disease:'MONDO:0005148',phenotype:'HP:0009124')` - or if not a constraint on the range (I am thinking of a regex pattern constraint), then another example on the `subject_id` slot so that folks know how to compose?

matentzn:
For now, I think we should not mandate a specific approach in SSSOM, because I don't want to close any doors to people that, for example, what to use crazy things like:

1. URL encoded OWL class expressions in RDF/XML format
2. URL encoded OTTR instantiations

Etc if they so must. I plan, however, explore how useful the URI expression language is to cover a variety of cases, and if it so holds up, and I see more than two distinct organisations implementing it, I might make a bid to lift it into the standard as a recommendation, with optional support by tools (sssom-java already supports it despite absolutely not having to do so).

I will probably implement some rudimentary support in sssom py which will work like this:

if m.subject_type == "composite entity expression"
      try:
          parse_json_url(curies.identifier(m.subject_id))
       except: ...

cmungall (approved):

PR #409, opened 2025-02-06
Add codespell support (config, workflow to detect/not fix) and make it fix few typos

yarikoptic:
More about codespell: https://github.com/codespell-project/codespell .

I personally introduced it to dozens if not hundreds of projects already and so far only positive feedback.

CI workflow has 'permissions' set only to 'read' so also should be safe.


project/jsonld/sssom_schema.jsonld
```diff
@@ -1844,7 +1844,7 @@
     {
       "name": "similarity_score",
       "definition_uri": "https://w3id.org/sssom/similarity_score",
-      "description": "A score between 0 and 1 to denote the similarity between two entities, where 1 denotes equivalence, and 0 denotes disjointness. The score is meant to be used in conjunction with the similarity_measure field, to document, for example, the lexical or semantic match of a matching algorithm.",
+      "description": "A score between 0 and 1 to denote the similarity between two entities, where 1 denotes equivalence, and 0 denotes disjointedness. The score is meant to be used in conjunction with the similarity_measure field, to document, for example, the lexical or semantic match of a matching algorithm.",
```
matentzn, line 1847: `disjointness` is a technical term in our domain, and should not be replaced.

project/jsonschema/sssom_schema.schema.json
```diff
@@ -350,7 +350,7 @@
                     "type": "string"
                 },
                 "similarity_score": {
-                    "description": "A score between 0 and 1 to denote the similarity between two entities, where 1 denotes equivalence, and 0 denotes disjointness. The score is meant to be used in conjunction with the similarity_measure field, to document, for example, the lexical or semantic match of a matching algorithm.",
+                    "description": "A score between 0 and 1 to denote the similarity between two entities, where 1 denotes equivalence, and 0 denotes disjointedness. The score is meant to be used in conjunction with the similarity_measure field, to document, for example, the lexical or semantic match of a matching algorithm.",
```
matentzn, line 353: See above.

project/shacl/sssom_schema.shacl.ttl
```diff
@@ -411,7 +411,7 @@ owl:Axiom a sh:NodeShape ;
             sh:order 28 ;
             sh:path pav:authoredOn ],
         [ sh:datatype xsd:double ;
-            sh:description "A score between 0 and 1 to denote the similarity between two entities, where 1 denotes equivalence, and 0 denotes disjointness. The score is meant to be used in conjunction with the similarity_measure field, to document, for example, the lexical or semantic match of a matching algorithm." ;
+            sh:description "A score between 0 and 1 to denote the similarity between two entities, where 1 denotes equivalence, and 0 denotes disjointedness. The score is meant to be used in conjunction with the similarity_measure field, to document, for example, the lexical or semantic match of a matching algorithm." ;
```
matentzn, line 414: See above.

project/sqlschema/sssom_schema.sql
```diff
@@ -46,7 +46,7 @@
 --     * Slot: mapping_date Description: The date the mapping was asserted. This is different from the date the mapping was published or compiled in a SSSOM file.
 --     * Slot: publication_date Description: The date the mapping was published. This is different from the date the mapping was asserted.
 --     * Slot: confidence Description: A score between 0 and 1 to denote the confidence or probability that the match is correct, where 1 denotes total confidence.
---     * Slot: similarity_score Description: A score between 0 and 1 to denote the similarity between two entities, where 1 denotes equivalence, and 0 denotes disjointness. The score is meant to be used in conjunction with the similarity_measure field, to document, for example, the lexical or semantic match of a matching algorithm.
```
matentzn, line 49: See above

pyproject.toml
```diff
@@ -25,3 +25,10 @@ build-backend = "poetry.core.masonry.api"
 
 [tool.poetry.extras]
 docs = ["linkml", "mkdocs-material"]
+
+[tool.codespell]
+# Ref: https://github.com/codespell-project/codespell#using-a-config-file
+skip = '.git*,*.pdf,*.lock'
+check-hidden = true
+ignore-regex = '\b(COMENT|EHR|LOD)\b'
```
matentzn, line 33: is this where to add exceptions?

matentzn (commented):
@yarikoptic Thank you so much for your contribution! This is very valuable. 

Just two comments / questions before we merge. Thank you!

Also, I am slightly concerned about having workflows blocking PRs for spelling mistakes but I will sort it out when the time comes.

Awesome!

matentzn:
@yarikoptic would you still like to proceed?

matentzn:
@yarikoptic Sorry for checking again on your valued contribution - will you be able to address the resulting issues?

pyproject.toml
yarikoptic: yes, if you would like "case-sensitivity" or some fancy regex. Otherwise, just uncomment/list in the line below.
Also url above points to how docs which also describe inline pragmas for "one of" hits, so you could just add to the line `codespell:ignore` in the comment and the line would be ignored.

project/jsonld/sssom_schema.jsonld
yarikoptic: indeed... I even considered to file fixup to codespell, but decided to not bother for now.  I will add a skip of this word. Thanks for spotting.

yarikoptic:
Thanks for the buzz and sorry that I missed prior one(s).  I have added to excludes, rebased and redone fixing. Should be all kosher now!

pyproject.toml
matentzn: Thank you!

matentzn (approved):

cmungall:
are you pinning to avoid spurious PR failures, e.g. https://github.com/linkml/linkml/issues/2536

yarikoptic:
commented there...  there is pinning only to `codespell-project/actions-codespell@v2` so it would stick to version 2 of codespell but use most recent version there thus could potentially indeed run into newly detected recently added to codespell dicts typos, or false positives (less likely imho but possible).

.github/workflows/codespell.yml
```diff
@@ -0,0 +1,25 @@
...
+    name: Check for spelling errors
+    runs-on: ubuntu-latest
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+      - name: Annotate locations with typos
+        uses: codespell-project/codespell-problem-matcher@v1
+      - name: Codespell
+        uses: codespell-project/actions-codespell@v2
```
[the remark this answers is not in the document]
gouttegd: Yes. There has been in v2.1 since then, but since the PR has been explicitly using v2, I looked up the commit ID for that tag.

matentzn (approved):

Issue #411, opened 2025-02-22
`similarity_measure`: Add as a metadata field


joeflack4:
## Overview
I have a mapping set where every entry in the set uses the same `simlarity_measure`. It'd be great to just put that information in the `metadata` instead of in each row.

I hear that this involves setting the slot to propagatable; instantiates: sssom:Propagatable.

I also hear that it needs to explicitly added to the slots of the MappingSet class.

## Additional info
Context: [Slack](https://obo-communitygroup.slack.com/archives/C0281J34Z6J/p1740256110704379)

PR #412, opened 2025-02-22
Allow using `similarity_measure` at the set level.

gouttegd:
Resolves [#411]

- ~~[ ] provide a full, working and valid example in `examples/`~~
- ~~[ ] provide a link to the related GitHub issue in the `see_also` field of the linkml model~~
- ~~[ ] provide a link to a valid example in the `see_also` field of the linkml model~~

This PR adds the `similarity_measure` slot to the `MappingSet` class, as a propagatable slot. It should be quite common that all `similarity_score` for all mappings in a set are determined using the same measure, so it makes sense to allow setting the measure once and for all at the set level, without having to repeat that information for all individual mappings.

### DIFF
--- a/src/sssom_schema/schema/sssom_schema.yaml
+++ b/src/sssom_schema/schema/sssom_schema.yaml
@@ -643,6 +643,8 @@ slots:
       To make processing this field as unambiguous as possible, we recommend using 
       wikidata CURIEs, but the type of this field is deliberately unspecified.
     range: string
+    annotations:
+      propagated: true
     examples:
       - value: https://www.wikidata.org/entity/Q865360
         description: the Wikidata IRI for the Jaccard index measure).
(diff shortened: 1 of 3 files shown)


src/sssom_schema/schema/sssom_schema.yaml
```diff
@@ -643,6 +643,8 @@ slots:
       To make processing this field as unambiguous as possible, we recommend using 
       wikidata CURIEs, but the type of this field is deliberately unspecified.
     range: string
+    annotations:
+      propagated: true
```
matentzn, line 647: We dont need to add 

```
instantiates: sssom:Propagatable
```
 
to the slot definition?

gouttegd: Weirdly enough, it seems to work (at least from the point of view of the code generator) even without it, but yes, I think that for correctness we need that.

matentzn (approved):
I tested a schema build, and it all works as intended.

ehartley (approved):
Looks good!

PR #432, opened 2025-04-09
Add link to sssom-js

nichtich:
The JavaScript library is in a very early state of development but already usable to convert SSSOM/TSV to JSON.


matentzn (approved):
Wow THAT is so cool! Thank you @nichtich!

Issue #439, opened 2025-05-01
[New metadata element]: sssom_version


gouttegd:
**Element id (e.g. creator_id, mapping_tool_version):**

```
sssom_version
```

**Value data type (e.g. URI, URL, text, xsd:boolean):**

```
xsd:string
```

**Description**

This is a set-level slot that identifies the version of the SSSOM specification that a mapping set declares itself to be compliant with. It MUST be a string of the form `X.Y`, where `X` is the specification’s major version number and `Y` is its minor version number. If that slot is absent, implementations MUST assume the set is compliant with version 1.0 of the specification.

(It should follow from that last statement that the `sssom_version` slot is MANDATORY for any set that is only compliant with a later version.)

**Complete example to a SSSOM file with this element**

```
# sssom_version: "1.1"
# curie_map:
#   HP: http://purl.obolibrary.org/obo/FBbt_
#   MP: http://purl.obolibrary.org/obo/UBERON_
# license: https://w3id.org/sssom/license/unspecified
# mapping_set_id: https://w3id.org/sssom/mappings/ac9e1878-73f4-4767-8402-a6c40e1b0835
```

(No Markdown table needed, this is a set-level metadata element only.)

**Rationale**

Now that SSSOM 1.0 has been published and that we are talking about amending the spec to add new slots (e.g. `source_mapping` in #437, `mapping_id` in #359, `mapping_set_confidence` in #438, …), it is time to think about how SSSOM is going to evolve without breaking existing implementations and existing mapping sets. This has been discussed already in #325, but in that discussion we merely outlined a general approach without any practical action.

Issue #325, opened 2023-10-24
Considerations on interoperability and evolution


gouttegd:
What are the plans, if any, to ensure the SSSOM TSV format remains interoperable

* across versions (a file produced with an implementation conforming to version _X_ can be used by an implementation conforming to version _X+1_),
* across implementations (a file produced by implementation _X_ — e.g. `sssom-py` — can be used another implementation _Y_ — e.g. `sssom-java`)?

Currently, this interoperability goal is explicitly _not_ sought after, as indicated both by the lower-than-one major version number (which, if we assume the project is using [semantic versioning](https://semver.org/), means that _anything_ can change _at any time_) and by the explicit warning on the top-level README: _Note that SSSOM is currently under development and subject to change._

With the SSSOM format being seemingly more and more used in the wild, I believe it is time to consider committing to some form of long-term stability of the format, and/or design ways to make the format evolve while preserving some basic interoperability across versions and implementations.

The following is a random set of ideas that could be explored. Feel free to discuss them, refute them, and add more.

## Defining a “core” set of metadata that will never change

We could select a handful of the most important mapping metadata and promote them to a “core set” that would be guaranteed never to change in any future evolution. This is similar to the “[minimal spec](https://github.com/mapping-commons/sssom/issues/260)” idea, though it could probably include slightly more metadata slots than the four mentioned in that ticket.

The principle here is that potential users could be confident that, no matter how the format evolves over time, as long as they only use the “core set” their files would always remain exploitable by any version of any conforming implementation. For users who need metadata from the “non-core set”, the situation would be the same as it is now for the entire standard: they would need to watch carefully the evolution of the standard to avoid being surprised by a breaking change.

For implementations, this would mean that they should only be strict when parsing the “core” metadata (at least by default — of course they can choose to allow users to specify a different behaviour). If they encounter a metadata slot they don’t recognise (because it’s an addition from a newer version of the spec), or a slot whose format has changed (e.g. because of a change such as the one envisioned [here](https://github.com/mapping-commons/sssom/issues/323), they may log a warning but should not fail altogether to parse the file.

## Allow each set to declare its own “must-understand” slots

This can be seen as a variation of the “core set” idea. Here, instead of having a fixed list of “core” metadata slots, the creators of a mapping set could define their own list of the slots they consider as critical.

[...]

PR #440, opened 2025-05-01
Add a `sssom_version` slot

gouttegd:
Resolves [#439]

This PR adds the new `sssom_version` set-level slot suggested in #439.

Contrary to the original suggestion of using a `xsd:string`-typed slot, it uses a enum-typed slot, with a new `sssom_version_enum` slot whose values represent the different versions of the specification (for now, the current 1.0 version and the upcoming 1.1). This should allow implementations to very easily check whether the version found in a set is one they support or not: if the value is not one of the “permissible values”, then either it’s a completely bogus value, or it’s one from a more recent version of the model than the one supported by the implementation.

It also add a new section in the documentation of the model, to explain how to deal with that `sssom_version` slot and, perhaps more importantly, to explain what are our promises regarding the evolution of the specification and backwards compatibility. That part may require further discussion before it is accepted, but it mostly follows the discussion that happened in #325. Basically, we promise never to break backwards compatibility across minor versions, which notably means that:

* a new minor version may only _add_ new slots or enum values, but never _remove_ a slot or an enum value;
* when adding a new slot in a minor version, that slot must be optional (possibly recommended, but never mandatory);
* a new minor version cannot make mandatory a slot that was previously optional.

Any change to the spec that would make invalid a set that was previously valid, MUST happen in a new _major_ version, for which we do not make any blanket promise of backwards compatibility.

Lastly, it adds a new meta model extension, _Versionable_, that can (and should) be used to mark any slot that has been added post-1.0.

### DIFF
--- a/src/docs/spec-model.md
+++ b/src/docs/spec-model.md
@@ -192,3 +192,47 @@ If a defined extension slot has a `type_hint` other than `http://www.w3.org/2001
 | http://www.w3.org/2001/XMLSchema#datetime | Implementations MAY check that the value is a date and time value in the ISO 8601 format (`yyyy-mm-ddThh:mm:ssTZ`) |
 
 Implementations MAY decide to recognise more types and to enforce type-specific constraints. For example, an implementation could recognise the type `http://www.w3.org/2001/XMLSchema#negativeInteger` and check that the value starts with a minus sign.
+
+## Versioning
+
+Starting from version 1.1 of the specification, the `MappingSet` class has an optional slot named `sssom_version` which indicates the version of the specification that the set declares itself to be compliant with.
+
+### Versioning rules
+
+The version number of the SSSOM specification has two components: a _major_ number _X_ and a _minor_ number Y, expressed as `X.Y`.
+
+A set that is compliant with a minor version _X.Y_ is also compliant with any minor version _X.Y+n_, for any value of _n_. The opposite is not true: a set compliant with a minor version _X.Y_ may not necessarily be compliant with a minor version _X.Y-n_.
+
+A set that is compliant with a major version _X_ may not be compliant with any other major version _X+n_ or _X-n_.
+
+Therefore, an implementation that is itself compliant with version _X.Y_ SHOULD always accept a set compliant with any version _X.Y-n_. It MAY reject outright a set compliant with any version _X.Y+n_ (more recent minor version), _X-n_ (older major version), or _X+n_ (more recent major version).
+
+In other words, the SSSOM specification guarantees backwards compatibility between two versions (in that a set compliant with an older version can be used with an implementation compliant with a newer version) only insofar as only the _minor_ version has changed.
+
+### Using the `sssom_version` slot
+
+When reading a SSSOM set:
+
+(A) If the set contains a `sssom_version` slot, implementations SHOULD check whether they recognize the indicated version as a supported version according to the rules in the previous section; if they don’t, they MAY reject the set outright.
+
+(B) If the set does not contain a `sssom_version` slot, it MUST be assumed to be compliant with version 1.0.
+
+When generating a SSSOM mapping set:
+
+(A) If the set uses slots or enum values that were added in more recent versions than 1.0, then the `sssom_version` slot MUST be set to the lowest version that defines all the slots effectively used.
+
+(B) If the set only uses slots or values that already existed in version 1.0, then the set is effectively compliant with said version 1.0 and the `sssom_version` slot MAY be omitted entirely.
+
+Note that, if the `sssom_version` slot is _not_ omitted, then it MUST be set to `1.1`, since that slot itself has been added in version 1.1. It follows that a `sssom_version=1.0` slot (a set that would declare itself to be compliant with version 1.0) is self-contradictory.
+
+### Model changes across versions
+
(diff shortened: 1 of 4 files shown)


matentzn (dismissed):

src/docs/spec-model.md
```diff
@@ -192,3 +192,47 @@ If a defined extension slot has a `type_hint` other than `http://www.w3.org/2001
...
+The version number of the SSSOM specification has two components: a _major_ number _X_ and a _minor_ number Y, expressed as `X.Y`.
+
+A set that is compliant with a minor version _X.Y_ is also compliant with any minor version _X.Y+n_, for any value of _n_. The opposite is not true: a set compliant with a minor version _X.Y_ may not necessarily be compliant with a minor version _X.Y-n_.
+
+A set that is compliant with a major version _X_ may not be compliant with any other major version _X+n_ or _X-n_.
+
+Therefore, an implementation that is itself compliant with version _X.Y_ SHOULD always accept a set compliant with any version _X.Y-n_. It MAY reject outright a set compliant with any version _X.Y+n_ (more recent minor version), _X-n_ (older major version), or _X+n_ (more recent major version).
+
+In other words, the SSSOM specification guarantees backwards compatibility between two versions (in that a set compliant with an older version can be used with an implementation compliant with a newer version) only insofar as only the _minor_ version has changed.
+
```
sierra-moxon, line 211: When reading this I thought you might be doing something different than semantic versioning...but after re-reading, I am not sure.  Why not just link the user to [semantic versioning docs](https://semver.org/)?  If you need to be more specific than plain semantic versioning, or want to clarify it, it would help me understand what you want me to do if I started from a place of understanding.

src/docs/spec-model.md
```diff
@@ -192,3 +192,47 @@ If a defined extension slot has a `type_hint` other than `http://www.w3.org/2001
...
+### Using the `sssom_version` slot
+
+When reading a SSSOM set:
+
+(A) If the set contains a `sssom_version` slot, implementations SHOULD check whether they recognize the indicated version as a supported version according to the rules in the previous section; if they don’t, they MAY reject the set outright.
+
+(B) If the set does not contain a `sssom_version` slot, it MUST be assumed to be compliant with version 1.0.
```
sierra-moxon, line 218: This seems like an important piece of info; any chance it can be decorated a bit more to stand out?

src/docs/spec-model.md
```diff
@@ -192,3 +192,47 @@ If a defined extension slot has a `type_hint` other than `http://www.w3.org/2001
...
+
+(A) If the set uses slots or enum values that were added in more recent versions than 1.0, then the `sssom_version` slot MUST be set to the lowest version that defines all the slots effectively used.
+
+(B) If the set only uses slots or values that already existed in version 1.0, then the set is effectively compliant with said version 1.0 and the `sssom_version` slot MAY be omitted entirely.
+
+Note that, if the `sssom_version` slot is _not_ omitted, then it MUST be set to `1.1`, since that slot itself has been added in version 1.1. It follows that a `sssom_version=1.0` slot (a set that would declare itself to be compliant with version 1.0) is self-contradictory.
+
+### Model changes across versions
+
+For all slots that were added to the specification after version 1.0, the LinkML model contains an `added_in` annotation that indicates the exact version in which the slot was introduced.
```
sierra-moxon, line 230: <3

sierra-moxon (approved):

src/docs/spec-model.md
gouttegd: This is basically the same “spirit” as semantic versioning, but the semantic versioning document is much too complex for what we need – for example, we do not want to deal with things like patch levels (3-component version numbers, e.g. `1.0.3`) or alpha or beta versions.

In effect, what we want is “semver but with only MAJOR.MINOR version numbers”.

Sure, I would expect any developer worth their salt to be able to deduce what we want from the semver specification, but given that this document is itself a specification, I’d rather be as precise as possible.

sierra-moxon: In a world where my dev time is limited (and salty 😄 ), learning a new method for something so familiar takes effort.  Are you sure you want me to use my effort to understand a new semantic versioning strategy especially with no hooks to get me familiar with the change you are trying to make?

e.g. "semver but with only MAJOR.MINOR version numbers" makes more sense to me (and sets the stage for me reading more about what you want to do to qualify semvar).

gouttegd: I can consider adding an introductory sentence like “versioning of the SSSOM mostly follows the Semantic Versioning specification but with MAJOR.MINOR version numbers. Details follow.” if that helps people that are already familiar with semver, but that’s about as far as I am willing to go.

In particular, I do not want developers already familiar to semdev to read that and think, “oh, OK, I already know that thing so I can pass“ – No! Any reference to semver  would be _informative_ only, the real spec is self-contained in this document.

src/docs/spec-model.md
gouttegd: This is a specification document. Almost _everything_ in it is important for anyone wanting to develop SSSOM-compliant software. This particular piece of info is already highlighted with a MUST keyword, which is the RFC 2119 keyword that denotes the highest level of importance (MUST = “absolute requirement”).

matentzn (approved):

PR #446, opened 2025-05-21
Add mapping_set_confidence slot

### LINKED ISSUE #438: [New metadata element]: mapping_set_confidence

```
mapping_set_confidence
```

```
xsd:float
```

**Description**

An optional mapping-set level confidence, assigned by the creator/producer of the set, based on whatever metric they deem fitting. This is different than an individual mapping confidence, such that it can be conveyed what's the overall confidence from the mapping provider in the correctness of the assertions.

Some examples:

- If an SSSOM file contains only human curations, then this could be the average of all confidences. Individual mappings can then have more granular confidence values, given by the curator of each row.
- If an SSSOM file contains lexical generations, then this annotation can correspond to a benchmarked confidence (i.e., accuracy) of the mapping tool. Individual mappings can have more granular confidence values, given by the tool for each generation

(This example can be given as a markdown table or a linked SSSOM file, feel free to edit the markdown table below)

```
# curie_map:
#   HP: http://purl.obolibrary.org/obo/FBbt_
#   MP: http://purl.obolibrary.org/obo/UBERON_
#   owl: http://www.w3.org/2002/07/owl#
#   rdf: http://www.w3.org/1999/02/22-rdf-syntax-ns#
#   rdfs: http://www.w3.org/2000/01/rdf-schema#
#   sssom: https://w3id.org/sssom/
# license: https://w3id.org/sssom/license/unspecified
# mapping_set_id: https://w3id.org/sssom/mappings/ac9e1878-73f4-4767-8402-a6c40e1b0835
# mapping_set_confidence: 0.8
```

| subject_id	| predicate_id	  | object_id	  | mapping_justification   | element_id_example	| 
| ----------- | --------------- | ----------- | ----------------------- | ------------------- |
| HP:0009124	| skos:exactMatch	| MP:0000003	| semapv:LexicalMatching	| YOUR EXAMPLE VALUE	| 
| HP:0008551	| skos:exactMatch	| MP:0000018	| semapv:LexicalMatching	| YOUR EXAMPLE VALUE	|

matentzn:
Resolves #438 

- [X] tests have been added/updated (if applicable)

This PR is adding "mapping_set_confidence" slot to the mapping set class.

### DIFF
--- a/examples/schema/mapping_set_confidence.sssom.tsv
+++ b/examples/schema/mapping_set_confidence.sssom.tsv
@@ -1,0 +1,15 @@
+#curie_map:
+#  HP: http://purl.obolibrary.org/obo/HP_
+#  MP: http://purl.obolibrary.org/obo/MP_
+#  orcid: https://orcid.org/
+#mapping_set_id: https://w3id.org/sssom/commons/examples/curie_map.sssom.tsv
+#mapping_set_confidence: 0.8
+#license: "https://creativecommons.org/publicdomain/zero/1.0/"
+#creator_id:
+#  - orcid:0000-0002-7356-1779
+#mapping_provider: "https://w3id.org/sssom/core_team"
+#mapping_tool: AgreementMakerLight
+#comment: This is an example file for the SSSOM for illustration only. The mapping_set_confidence value expresses the confidence of the creator of the mapping into the agent that produced the mappings, i.e. AgreementMakerLight.
+subject_id	predicate_id	object_id	mapping_justification
+HP:0009124	skos:exactMatch	MP:0000003	semapv:LexicalMatching
+HP:0008551	skos:exactMatch	MP:0000018	semapv:LexicalMatching
--- a/src/sssom_schema/schema/sssom_schema.yaml
+++ b/src/sssom_schema/schema/sssom_schema.yaml
@@ -116,6 +116,11 @@ slots:
   registry_confidence:
     description: This value is set by the registry that indexes the mapping set. It reflects the confidence the registry has in the correctness of the mappings in the mapping set.
     range: double
+    minimum_value: 0.0
+    maximum_value: 1.0
+    examples:
+      - value: 0.95
+        description: A confidence score of 0.95, indicating 95% confidence.
   last_updated:
     description: The date this reference was last updated.
     range: date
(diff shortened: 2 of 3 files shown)


examples/schema/mapping_set_confidence.sssom.tsv
```diff
@@ -0,0 +1,15 @@
+#curie_map:
```
gouttegd, line 1: Given that the `mapping_set_confidence` slot did not exist in SSSOM 1.0 and will be an addition in SSSOM 1.1, this set is only compliant with version 1.1 and therefore should have a `sssom_version: "1.1"` slot.

src/sssom_schema/schema/sssom_schema.yaml
```diff
@@ -356,6 +361,21 @@ slots:
     slot_uri: dcterms:description
     examples:
       - value: "This mapping set was produced to integrate human and mouse phenotype data at the IMPC. It is primarily used for making mouse phenotypes searchable by human synonyms at https://mousephenotype.org/."
+  mapping_set_confidence:
```
gouttegd, line 364: Since this will be a new slot in SSSOM 1.1, please add:

```yaml
mapping_set_confidence:
  instantiates: sssom:Versionable
  annotations:
    added_in: "1.1"
```

to properly mark the slot as compliant with SSSOM 1.1+.

gouttegd (changes requested):
Looks good to me, but please mark the slot as being a new slot in SSSOM 1.1.

(This should be a general requirement from now on for all slot additions.)

examples/schema/mapping_set_confidence.sssom.tsv
cthoyt: do we not validate that extraneous fields aren't present?

cthoyt (approved):

examples/schema/mapping_set_confidence.sssom.tsv
gouttegd: Depends on whether undefined extensions are supported or not.

Without support for undefined extensions, non-standard slots are _ignored_ (possibly with a warning, though that’s not mandatory). That’s what would happen with SSSOM-Py (which does not support extensions at all), and with SSSOM-Java by default (where support for undefined extensions must be explicitly enabled).

With support for undefined extensions, any non-standard slot is treated as an undefined extension.

In any case, the point here is that if you want a SSSOM 1.1 slot (like this `mapping_confidence_set`) to be recognised as a valid slot instead of being ignored or treated as an undefined extension, the set _must_ declares itself as a 1.1 set, by means of the `sssom_version="1.1"` slot.

matentzn: Thank you @gouttegd I added the slot information!

src/sssom_schema/schema/sssom_schema.yaml
matentzn: Thanks, done! I need to get used to that!

cthoyt (approved):

gouttegd (approved):

Issue #344, opened 2024-01-16
How to associate labels to ids in repeated slots


dr-shorthair:
Some properties of Mappings have unrestricted cardinality, and some IDs and matching labels (e.g. `author_id`+`author_label`, `reviewer_id`+`reviewer_label`). 

Is there a convention to associate these correctly?

matentzn:
@dr-shorthair Thank you for reaching out! 

There is currently no way to associate author_id and author_label - and this is by design. From the early days of SSSOM, we decided that the core design feature of SSSOM is simplicity, which meant, in essence, avoiding a nested data model. There are no nested elements in sssom at all - formally, all elements are associated directly with the mapping. What you would want (and many others like you) is to be able to have, in the datamodel at least, an `author` element with two slots, `id` and `label`. Unfortunately, this is not supported, and won't be in the standard model. There have been debates on a separate more structured model, but so far no one has shown enough interest to spearhead the design of such a model (which could be quite complex).

I personally think of it like this:

We should always use `author_id` if possible. Only if not possible, use `author_label`. Managing metadata about people involved in the mapping process should be managed elsewhere (else, why not add "author_expertise" or "author_affiliation" to the metadatamodel?).

In any case, nothing here is clear cut, and I can see a lot of push back against what I am saying :P

dr-shorthair:
Thanks. In my original question, I almost commented that *_label should be treated as non-normative, with *_id being the default, but wasn't quite sure how to phrase that.

gouttegd:
Re-opening because @matentzn ’s answer, about how `author_id` and `author_label` (and other similar slots such as `creator_id`, etc.) are _not_ linked and should _not_ be thought of as being linked, should really be part of the spec. The question in the ticket is a valid one (I have been asking that myself), and the spec should provide an answer to it.

Right now the spec about `author_label` simply says: “In the spirit of provenance, consider using author_id instead.” That is not enough.

PR #451, opened 2025-05-28
Clarify the absence of link between author_id and author_label.

gouttegd:
Resolves [#344]

- ~~[ ] `docs/` have been added/updated if necessary~~ not needed, doc is in the model
- ~~[ ] tests have been added/updated (if applicable)~~ nothing to test
- ~~[ ] [CHANGELOG.md](https://github.com/mapping-commons/sssom/blob/master/CHANGELOG.md) has been updated.~~ not needed, this is a clarification only

There is not supposed to be any link between `author_id` and `author_label`. In particular, `author_label[n]` is _not_ intended to be the human-friendly label of `author_id[n]`.

This has seemingly always been the intention (see notably this comment: https://github.com/mapping-commons/sssom/issues/344#issuecomment-1893588484) but has never been made explicit in the schema, which is what this PR is about.

Likewise for `reviewer_id`/`reviewer_label` and `creator_id`/`creator_label`.

### DIFF
--- a/src/sssom_schema/schema/sssom_schema.yaml
+++ b/src/sssom_schema/schema/sssom_schema.yaml
@@ -395,8 +395,12 @@ slots:
       - value: orcid:0000-0002-7356-1779
         description: The ORCID of the creator of the mapping.
   creator_label:
-    description: A string identifying the creator of this mapping. In the spirit of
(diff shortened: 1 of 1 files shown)


matentzn (approved):
Yes, thank you!

Issue #359, opened 2024-04-09
New slot mapping_id


nichtich:
I bet this has been mentioned before but I did not find the corresponding issue: SSSOM lacks a slot to identify individual rows. This is required to fully map from/to JSKOS Mapping format (#249) but I'm sure there are other use cases as well.

nichtich:
Looks like `[see_also](https://mapping-commons.github.io/sssom/see_also/)` can be used to hold the URI of the individual mapping, can't it?

matentzn:
I renamed the issue now, and I keep stumbleing over the fact that this issue has been discussed a lot of times, in the context of other issues like #91. 

Adding `mapping_id` as a required slot is out of the question I think, because identifier management is too much churn for most users that just want to add a quick table with mappings to their repo.

I see three options:

1. We create an _optional_ slot _mapping_id_. This will limit its use in other contexts, like provenance focused properties ("mapping A was derived from a combination of mapping B and C").
2. We create a _non-curated_ slot _mapping_id_. Basically, we create a defined protocol in which the "mapping" part of a SSSOM row can be turned into an identifier, for example:
    - concatenate subject_id, predicate_id, predicate_modifier and object_id (and potentially mapping_set_id for context) into a pipe-separated string
    - do the above, and base64 encode
    - etc, etc
    - The advantage of this solution is that we can safely construct identifiers even where no `mapping_id` exists.
 3. A combination of the above?

I dont know the answer, but we need to make a decision, so thanks for bringing this up again. There are so many advantages of having mapping_ids including:

1. You dont need ugly re-fication with blank nodes in the RDF serialisation
2. You can describe provenance like: this and that mapping were used to derive this mapping
3. You can create lookup URLs on Mapping browsers like OxO for additional metadata.
4. You can link the mapping_id in an ontology as mapping metadata instead of adding the 30 SSSOM metadata elements and keep things very separate
5. its more FAIR (everything should have a PID).

gouttegd:
Let’s first clarify _what_ needs to be identified here: _mappings_ proper, or _mapping records_?

For clarity, what I call a *mapping proper* (or *core mapping*) is purely a triple {subject, predicate, object} (and optionally a predicate modifier). A _mapping record_ is a mapping proper plus additional metadata (justification, provenance, etc.) – basically, a mapping record is a row in a SSSOM/TSV file.

SSSOM primarily deals with *mapping records*, _not_ *core mappings*.

[...]

Issue #436, opened 2025-04-14
Define a standard hashing procedure


gouttegd:
To allow implementations to _unambiguously_ and _consistently_ assess whether two mapping records are identical, the spec could/should define a standard hashing function and how the hash should be computed.

Considerations:

**A. Choice of the hashing function.** Not really important, we just need to pick one. Easiest option would probably be SHA-256, which is almost ubiquitously available in all programming languages. We do not really need its cryptographic properties of collision resistance and preimage resistance, but they don’t hurt.

**B. What to hash?** Simply put: _everything_. That is, _all_ the slots that make up a mapping record. This should also probably include any non-standard slot.

**C. How to hash?** _This_ is the real question. We need to define a serialisation format such that any given mapping record can have one, **and only one**, possible serialised form. The “canonical SSSOM/TSV” format as currently defined in the spec is not suitable, as it still leaves some room for variations across implementations.

One option would be to serialise the record into a [canonical S-expression](https://en.wikipedia.org/wiki/Canonical_S-expressions), e.g.

```
(7:mapping((10:subject_id44:http://purl.obolibrary.org/obo/FBbt_00001234)(12:predicate_id:46http://www.w3.org/2004/02/skos/core#exactMatch)(9:object_id45:http://purl.obolibrary.org/obo/UBERON_0005678)(21:mapping_justification51:https://w3id.org/semapv/vocab/ManualMappingCuration)(10:creator_id(37:https://orcid.org/0000-0000-1234-567837:https://orcid.org/0000-0000-5678-1234))
```

Regardless of the exact serialisation format, prior to serialisation and hashing: (1) all CURIEs must be expanded to their full-length form; (2) all propagatable slots must be propagated; (3) all multi-valued slots must be sorted.

gouttegd:
⬆ @matentzn FYI, follow-up to our discussion today.

matentzn:
Yes this captures it greatly, and S-expression is a good idea. I think this is what James Overton also used in one of his earlier tools to represent OWL expressions in a canonical form. Thanks for writing this up!

cthoyt:
I have implemented my own hashes in SeMRA, but I would be much happier for this to be an upstream discussion in the SSSOM spec! Happy to help on this.

gouttegd:
This is what the standard hashing procedure could look like if we specify it based on how hashing is currently done in SSSOM-Java:

## Step 0: Propagate all condensed slots

[...]

PR #452, opened 2025-05-31
Add `record_id` slot.

gouttegd:
Resolves [#359]

- ~~[ ] `docs/` have been added/updated if necessary~~ documentation is embedded in the LinkML schema.

- [x] make sure any new slot is annotated with the appropriate `added_in` annotation

This PR adds a new slot to the `Mapping` class, `record_id`, intended to hold a unique identifier for a given mapping.

The slot is optional, so as not to break compatibility with existing SSSOM 1.0 sets. This also means that, while we can define the slot as a “unique key” for the Mapping class, we _cannot_ define it as the “identifier”, because in LinkML identifier slots are automatically mandatory.

The identifier is intended to be completely opaque. How to generate identifiers is left to the producers of SSSOM sets, and no meaning of any sort should be assigned to an identifier.

### DIFF
--- a/examples/schema/record-ids.sssom.tsv
+++ b/examples/schema/record-ids.sssom.tsv
@@ -1,0 +1,10 @@
+# sssom_version: "1.1"
+# curie_map:
+#   HP: http://purl.obolibrary.org/obo/FBbt_
+#   MP: http://purl.obolibrary.org/obo/UBERON_
+#   RI: https://example.org/sets/record-id#
+# mapping_set_id: https://example.org/sets/record-id
+# license: https://creativecommons.org/publicdomain/zero/1.0/
+record_id	subject_id	predicate_id	object_id	mapping_justification
+RI:0000001	HP:0009124	skos:exactMatch	MP:0000003	semapv:LexicalMatching
+RI:0000002	HP:0008551	skos:exactMatch	MP:0000018	semapv:LexicalMatching
--- a/src/sssom_schema/schema/sssom_schema.yaml
+++ b/src/sssom_schema/schema/sssom_schema.yaml
@@ -779,6 +779,15 @@ slots:
     see_also:
       - https://github.com/mapping-commons/sssom/issues/328
       - https://github.com/mapping-commons/sssom/blob/master/examples/schema/extension-slots.sssom.tsv
+  record_id:
+    description: A unique identifier for a mapping record.
+    range: uriorcurie
+    instantiates: sssom:Versionable
+    annotations:
+      added_in: "1.1"
+    see_also:
+      - https://github.com/mapping-commons/sssom/issues/359
+      - https://github.com/mapping-commons/blob/master/examples/schema/record-ids.sssom.tsv
 classes:
   mapping set:
     description: Represents a set of mappings.
@@ -823,6 +832,7 @@ classes:
   mapping:
     description: Represents an individual mapping between a pair of entities.
     slots:
+    - record_id
     - subject_id
     - subject_label
     - subject_category
@@ -869,6 +879,17 @@ classes:
     - other
     - comment
     class_uri: owl:Axiom
+    unique_keys:
+      record_identifier:
+        description: >-
+          Each mapping within a mapping set MAY be identified by a unique, opaque record
+          identifier. This slot MUST be used consistently, in that either all mappings in
+          the set have a such a record identifier, or none of them have one. The
+          behaviour when a set contains both mappings with a record identifier and
+          mappings without a record identifier is unspecified. The behaviour when two
+          mappings have the same record identifier is unspecified.
+        unique_key_slots:
+          - record_id
     rules:
       - preconditions:
           slot_conditions:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@
 - Add `sssom_version` slot to the `MappingSet` class ([issue](https://github.com/mapping-commons/sssom/issues/439)).
 - Change the type of the `see_also` slot to `xsd:anyURI` ([issue](https://github.com/mapping-commons/sssom/issues/422)).
 - Add `mappings_set_confidence` slot to the `MappingSet` class ([issue](https://github.com/mapping-commons/sssom/issues/438)).
+- Add `record_id` slot to the `Mapping` class ([issue](https://github.com/mapping-commons/sssom/issues/359)).
 - TBD
 
 ## SSSOM version 1.0.0


src/sssom_schema/schema/sssom_schema.yaml
```diff
@@ -779,6 +779,15 @@ slots:
     see_also:
       - https://github.com/mapping-commons/sssom/issues/328
       - https://github.com/mapping-commons/sssom/blob/master/examples/schema/extension-slots.sssom.tsv
+  record_id:
+    description: A unique identifier for a mapping record.
```
matentzn: There should be some level of clarity I think here whet the identifier identifies:

1. A _version_ of a mapping record
2. A conceptual mapping record

So if we have for example the same exact record but one with an extension slot (or a comment), should it still be considered "the same"?

matentzn: And, if the record is identical, but the mapping set as a whole is different (including one or more propagatable slots), what then?

gouttegd: The record identifier identifies a _record_. I am not sure it can be more “clear” than that.

SSSOM has _only_ a concept of “record”, represented by the `Mapping` class. There is _no_ concept of “version of a record” or of “conceptual mapping record”. A record is basically a row in a SSSOM/TSV file – in fact I could have called the new slot `row_id`, which may have been more explicit but I don’t like the idea of linking the slot name to one particular serialisation format.

So if you have two _rows_ in a set, you have two different records, and they MUST have a different record identifier. It doesn’t matter that they are almost identical and differ only by one slot.

gouttegd: Now we could _also_ introduce another slot to allow us to represent “versions of a record”, but that is another issue.

For example, we could have something like this:

```
record_id  mapping_id  subject_id  predicate_id     object_id  mapping_justification         status
REC:00001  MAP:0001    HP:1234     skos:exactMatch  DO:5678    semapv:ManualMappingCuration  status:introduced
REC:00002  MAP:0001    HP:1234     skos:exactMatch  DO:5678    semapv:ManualMappingCuration  status:reviewed
```

where we have two records (REC:0001, REC:0002) that represent different versions of the same “conceptual mapping” MAP:0001. But as I said, this is a different issue, and one that we can tackle later – in particular, we do _not_ need this `mapping_id` slot for RDF serialisation, because this slot does _not_ identify a record, it only serves to link several records together, to mark them as being about the same mapping.

gouttegd: > And, if the record is identical, but the mapping set as a whole is different

If you have two records in two different sets, you still have _two_ records, and the mere fact that they are in two different sets should be enough to say that the records themselves are different.

Record identifiers are opaque and SSSOM producers can generate them the way they want, but it would probably be a good idea to make them using the mapping set ID as a base.

That is, if you have a mapping set with a `mapping_set_id` of `https://example.org/myset`, then records within that set could be something like `https://example.org/myset#0001`, `https://example.org/myset#0002`, etc.

Mapping set ID-dependent record identifiers would ensure that you can never have clashing record IDs when you merge mapping sets together.

examples/schema/record-ids.sssom.tsv
```diff
@@ -0,0 +1,10 @@
+# sssom_version: "1.1"
+# curie_map:
+#   HP: http://purl.obolibrary.org/obo/FBbt_
+#   MP: http://purl.obolibrary.org/obo/UBERON_
+#   RI: https://example.org/sets/record-id#
+# mapping_set_id: https://example.org/sets/record-id
+# license: https://creativecommons.org/publicdomain/zero/1.0/
+record_id	subject_id	predicate_id	object_id	mapping_justification
```
cthoyt, line 8: can you please give more examples of places where the record_id should be the same, and when it should be different? it's not clear to me which fields in the mapping/mapping set should be considered for this

gouttegd: It’s a _unique identifier_. There are _no_ “places where the record_id should be the same”. Two records should always have a different identifier.

cthoyt: I'm still having a hard time understanding what a record is - I added a bit more text in the body of the PR about this

gouttegd: Point taken. I _thought_ I had introduced the notion of “[mapping] record” (as in, in instance of the `Mapping` class) back when I overhauled the documentation last year, but obviously I had not and this still has to be done.

gouttegd:
> Do you want record_id to be universally unique, or just locally unique to the mapping set?

For the spec, what matters is uniqueness within a mapping set.

I expect many users will want to use globally unique identifiers, and they can absolutely do so, but I believe the spec should not enforce that. At most, we can promote that as a “good practice” in the _documentation_ section of the SSSOM website, but this out of scope for the _specification_.

> For maximum linkability and discoverability I'd like to request the format of the identifier be specified as + '/' + fragment. I'm OK with the fragment being a random (IRI-compatible) string, but again, if you're going to the trouble of creating an identifier field, let's standardize it appropriately the first time. Web page anchors and semantic artifacts have taught us some useful lessons.

Flatly disagree. The spec should not mandate anything about the “format” of the identifier. As far as the spec is concerned, identifiers have no “format”, they are opaque strings. Any attempt at enforcing a particular format will only reduce the flexibility and the usefulness of the format.

(For an example of what I mean: it’s because we treat identifiers as merely opaque string without mandating any particular format that an idea such as the “[URI Expression language](https://github.com/monarch-initiative/uri-expression-language)” is even possible; the idea would have been dead in the water if someone had decided that semantic identifiers had to follow a specific format.)

> Is an identical record (ignoring the record_id) in the same mapping set legal?

It is not explicitly forbidden, so it is legal, yes.

>  If so, and if it has a record_id, it would only be legal if the record_id does not match its nominal twin, right?

As currently written in the present PR, yes. More precisely, 2 records in the same set with the same `record_id` is a _undefined behaviour_ scenario – that is, we don’t tell implementations what they should do in that case. They can reject the set outright, they can drop the first record, drop the second record, attempt to merge the records… it’s up to them.

I am open to the idea of defining an exact behaviour to adopt (and to mandate in the spec) in that scenario, _if_ we can agree on what the correct behaviour would be.

> Re identifiers for the triple vs the whole record: It may be helpful to consider that if you have an identifier for the whole record, it is trivial to support a variant which is identifying the triple within that record.

I am on the record (pun intended) being opposed to the idea of an identifier for the subject-predicate-object triple (let’s call that the “SPO ID” here):

* this creates more failure modes: what if two records have the same SPO ID but a different SPO triple? or the same SPO triple but a different SPO ID?
* the SPO can _already_ be uniquely identified by the combination of the identifiers of its three components;
* this is neither (1) what has been requested in #359 (and needed for mapping to JSKOS) nor (2) what we need for RDF serialisation.

In any case, this is orthogonal to the idea of an identifier for an entire record. If people _really_ want a SPO ID, they can open a new issue and make a case for it. But for now [I have yet to see _anyone_](https://github.com/mapping-commons/sssom/discussions/360) make a real case for SPO IDs.

> I really wish record_id could be a formal identifier, as in required. If that breaks previous schemas then call it a breaking schema and support both.

SSSOM has been in used for several years already and until now the lack of a required record identifier has not been seen as a problem – or at least, not problematic enough to motivate people to do something about it.

I am not flatly opposed to the idea of making `record_id` mandatory for SSSOM 1.1+ sets (provided we mandate that implementations SHOULD still support ID-less 1.0 sets), but I’d like to hear more from users before making such a drastic change. I know _my_ use cases for SSSOM certainly do not require a record identifier, and in fact SSSOM would be slightly more cumbersome to use if such identifiers were mandatory.

> Umm, there is a credible argument for including a checksum (validation) information in the identifier. Is that also included in your objection?

Yes. For at least two reasons (I’m pretty sure I already exposed them in #359):

* With a content-dependent identifier, any change to the contents of the record changes the identifier, which means the identifier is not stable (you fixed a typo somewhere? your record ID has changed; oh, you already used that record ID to refer to that particular record in your database? well, I guess you need to go update your database then) – yes, there may be cases where this is actually a desired behaviour (see for example Git commit IDs), but I seriously doubt this would _always_ be desired.
* If the identifier is derived in any way from the contents of the record (e.g., through some kind of hashing), it means that it cannot possibly be manually created by the human editing the mapping set. Any creation or modification of a SSSOM/TSV file would require passing the file through an ID-generating step to ensure that all IDs are correctly generated. This breaks an important promise of SSSOM, namely that SSSOM/TSV files are _simple_ files that do not require editors to use specialised tool.

> In the end I'm fine with explicitly avoiding specifying the scheme for the record_id. But you know people will make the record_id depend on the content if you don't disallow that.

Yes! That’s the whole point of _not_ specifying any scheme! For the spec, the identifier is an opaque string and nothing more.

If some users _want_ to use content-derived IDs, they can absolutely do so. The spec will not get in their way. In fact, the spec could even define a standard algorithm to generate such a content-derived ID (the idea has been floated here #436) that implementations could then provide to the users. I have nothing against that at all. I am simply flatly opposed to _mandating_ the use of such IDs.

> Re versioning, I am pretty sure the system will be much more robust if identifiers do not embed versioning information. That information about a record can be (and is) handled by another attribute.

Fully agree.

gouttegd:
> As far as the spec is concerned, identifiers have no “format”, they are opaque strings.

Let me correct that as it’s not quite true: we mandate at least that identifiers must be URIs, since they are typed as EntityReference.

That’s already an important deviation from the idea of completely “opaque” strings, and I am not willing to deviate anymore than that by mandating on top of it a certain format of URI.

matentzn:
Great discussion all - as always it is a bit hard to follow complex intertwined discussions on the main body of the pull request; Maybe in the future we should always move discussions back to the issue discussion, especially for big changes like this one.

I only want to add two thoughts to this PR (beyond what was already suggested):

1. To reduce the scope of this PR to the barebones spec part (identifier format, i.e. entity reference, examples), and separate best practices into a separate PR (global uniqueness, mapping set id as base, etc). I agree with most of @graybeal suggestion on promoting best practice, but I also fully agree with @gouttegd general sentiment to keep such considerations out of the spec.
2. I am firmly against the _requirement_ to _asserting_ a record_id in SSSOM, either now, or in the future. ID management is a big burden for anyone producing a mapping set, and the main, original community for which SSSOM is designed is not the interoperability nerds like we are, but the **scientists who need to publish a mapping as part of their work/publication in a table format**. However, I would be open to a "standardised" ID generation protocol for the RDF-based serialisations. (I have no opinion on content-derived ids but would find it convenient if there was a way for two people generating record_ids for the same exact mapping set to end up having the same exact record_id in RDF (not relevant to PR)).

@gouttegd there is a lot of talk in this PR - I am not sure from your responses which you are planning to address.

gouttegd:
As for @graybeal comments:

It boils down to whether we agree that the spec should _not_ be prescriptive about the format of identifiers and how they are generated. If we do, then there’s nothing else to say.

If we don’t, then again we might as well close this PR and postpone the creation of a record ID slot (and therefore also postpone specifying the RDF serialisation) until we agree on what SSSOM record identifiers should or should not be.

gouttegd:
I’d also like to point out that the PR is merely a direct implementation of what I proposed [more than a year ago](https://github.com/mapping-commons/sssom/issues/359#issuecomment-2047144651) (modulo the name, which I changed from `mapping_id` to `reord_id`) as a solution to #359. At the time and in the following 13 months nobody raised an objection to it to that particular proposal, instead the discussion got bogged down on issues about content-derived identifiers or core-mapping-id vs record-id.

gouttegd:
> I have no opinion on content-derived ids but would find it convenient if there was a way for two people generating record_ids for the same exact mapping set to end up having the same exact record_id in RDF

I guess we can discuss that when we come to discussing the RDF spec, but you are hereby warned that I will only agree with that if it is a mere _possibility_, not a _requirement_.

The resource identifier for the RDF representation of a mapping record should be the value of the `record_id` slot and that’s it. Independently of how that `record_id` has been generated.

Now if a set has _no_ `record_id` slots, I don’t mind suggesting that implementations MAY auto-generate `record_id` values by deriving them from the records (though I’d much rather suggest doing something like `mapping_set_id` + incremental serial number).

But if a set does have `record_id` slots, I am flatly opposed to mandating that the RDF serialisation should discard them and use its own auto-generated IDs instead.

src/sssom_schema/schema/sssom_schema.yaml
```diff
@@ -779,6 +779,26 @@ slots:
     see_also:
       - https://github.com/mapping-commons/sssom/issues/328
       - https://github.com/mapping-commons/sssom/blob/master/examples/schema/extension-slots.sssom.tsv
+  record_id:
+    description: >-
+      A unique identifier for a mapping record, that is for an instance of the Mapping
+      class (in the SSSOM/TSV serialisation, this corresponds to an individual row).
```
matentzn:
```suggestion
      class (in the SSSOM/TSV serialisation, this corresponds to an individual row after propagation is applied).
```

matentzn (dismissed):
I am happy with this proposal. To make me _perfectly_ happy I would like to see a sentence that clarifies that we interpret slot condensing as syntactic sugar, and that a condensed slot at mapping_set level still belongs to the record (while a mapping set level slot in general does not), but maybe this does not belong in the spec.

@cthoyt what do you think? IMO the purpose of this proposal is _mostly_ to enable a standard way to generate resource IDs in RDF, and I hope (expect) to not see these record ids so much in the wild. I think this is good.

src/sssom_schema/schema/sssom_schema.yaml
gouttegd: > I would like to see a sentence that clarifies that we interpret slot condensing as syntactic sugar

I think this is not the place to do so. Since we seem to agree on using the term “record” I was planning (as I already announced in a previous comment) another independent PR to formally introduce the notion of a “record“, which would be the right place for this kind of clarification.

But (1) people have been requested a record identifier slot for more than one year and (2) we really need such a slot before we can finally tackle the issue of RDF serialisation, so if injecting this kind of out-of-place clarification is all it takes to make the record identifier slot find its way into the model, then fine, let’s do it.

matentzn:
> Not generate. Store.

Haha my bad. I do understand this proposal is about storing, not about generating. Bad choice of words

matentzn (dismissed):

examples/schema/record-ids.sssom.tsv
```diff
@@ -0,0 +1,10 @@
...
+RI:0000001	HP:0009124	skos:exactMatch	MP:0000003	semapv:LexicalMatching
+RI:0000002	HP:0008551	skos:exactMatch	MP:0000018	semapv:LexicalMatching
```
cthoyt, line 10: demonstrate that the same core triple might appear in multiple records, which each get unique record IDs

```suggestion
RI:0000002	HP:0008551	skos:exactMatch	MP:0000018	semapv:LexicalMatching
RI:0000003	HP:0008551	skos:exactMatch	MP:0000018	semapv:ManualMappingCuration
```

cthoyt (approved):
I am happy with the improvement of explanation, but would still like to see more human-level documentation be part of proposals beyond the gobbledygook that comprises changing the LinkML schema.

Since @gouttegd said he would follow up with this, I think you can accept my suggestion to improve the example file then finish this

cthoyt (approved):

Issue #448, opened 2025-05-23
Change definition of type Uri


nichtich:
The type [Uri](https://mapping-commons.github.io/sssom/Uri/) is underspecified ans misleading. It's description is "a complete URI" so one could assume an absolute URI (no relative URI reference) as defined in [RFC 3986] (URI). Then there is a link to http://www.w3.org/2001/XMLSchema#anyURI, referencing [xsd:anyURI in XML Schema (2004)](https://www.w3.org/TR/xmlschema-2/#anyURI). What is the normative definition of type Uri in SSSOM?

- a "complete" URI (as defined in [RFC 3986])
- any value allowed XML Schema (2004) data type `xsd:anyURI`
- something else (e.g. IRI)
- ...

Note that `xsd:anyURI` in XML Schema is *more* than absolute and relative URIs as defined in [RFC 3986] but it also support its own [escaping rules](https://www.w3.org/TR/2001/REC-xlink-20010627/#link-locators) to allow more characters.

I would prefer to refer to [RFC 3986] with absolute URI in the SSSOM data model only because in the end everything (relative URI, xsd:anyURI, IRI references...) is mapped to an absolute URI. The encoding of URIs in SSSOM/TSV *may* allow something more, but this needs to be defined more strictly. In theory, `xsd:anyURI` from [xsd:anyURI in XML Schema 1.1 (2012)](https://www.w3.org/TR/xmlschema11-2/#anyURI) or [RFC 3987] (IRI) could do but they come with several specific rules how to map its value to URI so just referencing these standards will likely not solve the issue unless the underlying specification has carefully been read.

[RFC 3986]: https://datatracker.ietf.org/doc/html/rfc3986
[RFC 3987]: https://datatracker.ietf.org/doc/html/rfc3987

matentzn:
I agree with the general proposal of further constraining the URI typed slots to be absolute.

@nichtich @gouttegd is the right way to:

Define a new type absolute_uri, which derives its specification from [RFC 3986](https://datatracker.ietf.org/doc/html/rfc3986):

```
types:
  absolute_uri:
    uri: xsd:anyURI
    base: str
    pattern: '^([a-zA-Z][a-zA-Z0-9+.-]*):(?:\\/\\/)?[^\s]*$'
```

And the use that type like this:

```
slots:
  license:
    range: absolute_uri
    description: "An absolute URI identifying the license for this mapping set."
```

nichtich:
[...]

PR #453, opened 2025-06-01
Clarify that URI-typed slots expect non-relative URI values.

gouttegd:
Resolves [#448]

URI-typed slots are expected to contain (non-relative) URIs only, but the LinkML `uri` type accepts both URIs _and_ relative URI references (something that is not very clear in the LinkML’s metamodel, but [the implementation of the `URI` type](https://github.com/linkml/linkml/issues/2726#issuecomment-2909331813) leaves no doubt that relative URI references are considered valid values).

So we add a new `NonRelativeURI` type, based on the LinkML `uri` type but with the added restriction that the URI must not be a relative URI reference. Of note, that restriction is only _documented_, not _enforced_: it is up to implementations to actually check that the value of a `NonRelativeURI`-typed slot is indeed a non-relative URI, as the underlying LinkML implementation will accept both non-relative and relative URI references.

All slots that were previously defined with a `uri` range are redefined to use the new `NonAbsoluteURI` type.

Since it has never been clear, in SSSOM 1.0, that URI-typed slots were supposed to contain only non-relative URIs, for backwards compatibility it is RECOMMENDED that implementations still accept relative URI references when processing sets compliant with version 1.0 of the spec.

### DIFF
--- a/src/docs/spec-model.md
+++ b/src/docs/spec-model.md
@@ -236,4 +236,4 @@ Not all changes can be annotated thusly in the LinkML model, though. For changes
 * The `similarity_measure` slot, which previously only existed on the `Mapping` class, has been added to the `MappingSet` class.
 * The value `composed entity expression` has been added to the `EntityType` enumeration.
 * The type of the `see_also` slot has been changed to `sssom:NonRelativeURI`. When parsing a SSSOM 1.0 set, implementations SHOULD accept arbitrary string values in that slot.
-* All slots that were typed as `xsd:anyURI` have been re-typed as `sssom:NonRelativeURI`. When parsing a SSSOM 1.0 set, implementations SHOULD accept relative URIs values in those slots.
+* All slots that were typed as `xsd:anyURI` have been re-typed as `sssom:NonRelativeURI`. When parsing a SSSOM 1.0 set, implementations SHOULD accept relative URI values in those slots.


src/sssom_schema/schema/sssom_schema.yaml
```diff
@@ -85,6 +85,9 @@ types:
     uri: rdfs:Resource
     see_also:
       - https://mapping-commons.github.io/sssom/spec/#tsv
+ AbsoluteURI:
+    typeof: uri
+    description: An absolute URI as per RFC 3986.
```
cthoyt: Could you please give a set of examples and counterexamples to make it clear what sort of things are being avoided here? 

Further, I think it would be nice to give the reader a quick summary of what RFC 3986 entails

src/sssom_schema/schema/sssom_schema.yaml
```diff
@@ -145,13 +177,13 @@ slots:
...
-    range: uri
+    range: NonRelativeURI
   documentation:
     description: A URL to the documentation of this mapping commons.
-    range: uri
+    range: NonRelativeURI
   homepage:
     description: A URL to a homepage of this mapping commons.
-    range: uri
+    range: NonRelativeURI
```
matentzn, line 193: Technically this case here is even stricter than `NonRelativeURI` as urns are NOT URLs?

src/sssom_schema/schema/sssom_schema.yaml
```diff
@@ -145,13 +177,13 @@ slots:
   imports:
     description: A list of registries that should be imported into this one.
     multivalued: true
-    range: uri
+    range: NonRelativeURI
   documentation:
     description: A URL to the documentation of this mapping commons.
-    range: uri
+    range: NonRelativeURI
```
matentzn, line 190: Technically this case here is even stricter than NonRelativeURI as urns are NOT URLs?

src/sssom_schema/schema/sssom_schema.yaml
```diff
@@ -145,13 +177,13 @@ slots:
   imports:
     description: A list of registries that should be imported into this one.
     multivalued: true
-    range: uri
+    range: NonRelativeURI
```
matentzn, line 187: Technically this case here is even stricter than NonRelativeURI as urns are NOT URLs?

src/docs/spec-model.md
```diff
@@ -235,4 +235,5 @@ Not all changes can be annotated thusly in the LinkML model, though. For changes
 
 * The `similarity_measure` slot, which previously only existed on the `Mapping` class, has been added to the `MappingSet` class.
 * The value `composed entity expression` has been added to the `EntityType` enumeration.
-* The type of the `see_also` slot has been changed to `xsd:anyURI`.
+* The type of the `see_also` slot has been changed to `sssom:AbsoluteURI`. When parsing a SSSOM 1.0 set, implementations SHOULD accept arbitrary string values in that slot.
```
matentzn:
```suggestion
* The type of the `see_also` slot has been changed to `sssom:NonRelativeURI`. When parsing a SSSOM 1.0 set, implementations SHOULD accept arbitrary string values in that slot.
```

matentzn (changes requested):

src/docs/spec-model.md
```diff
@@ -235,4 +235,5 @@ Not all changes can be annotated thusly in the LinkML model, though. For changes
 
 * The `similarity_measure` slot, which previously only existed on the `Mapping` class, has been added to the `MappingSet` class.
 * The value `composed entity expression` has been added to the `EntityType` enumeration.
-* The type of the `see_also` slot has been changed to `xsd:anyURI`.
+* The type of the `see_also` slot has been changed to `sssom:NonRelativeURI`. When parsing a SSSOM 1.0 set, implementations SHOULD accept arbitrary string values in that slot.
+* All slots that were typed as `xsd:anyURI` have been re-typed as `sssom:NonRelativeURI`. When parsing a SSSOM 1.0 set, implementations SHOULD accept relative URIs values in those slots.
```
ehartley:
```suggestion
* All slots that were typed as `xsd:anyURI` have been re-typed as `sssom:NonRelativeURI`. When parsing a SSSOM 1.0 set, implementations SHOULD accept relative URI values in those slots.
```

ehartley (changes requested):
Should the range of the `prefix_url` slot also be changed to `NonRelativeURI`?

gouttegd:
> Should the range of the prefix_url slot also be changed to NonRelativeURI?

I sure hope no one would ever even think of using relative URI references in a curie map! What would that even mean? :D 

… _But_ if we don’t explicitly forbid it, then it is almost guaranteed that someone, somewhere, will do exactly that just because they (think they) can, so I agree. I will change the range of `prefix_url` now.

(As an aside, it is unfortunate that this slot has been named `prefix_urL`, as `prefix_urI` would have been better. Well, too late for that now.)

gouttegd:
> I sure hope no one would ever even think of using relative URI references in a curie map! What would that even mean? :D
>
> … But if we don’t explicitly forbid it, then it is almost guaranteed that someone, somewhere, will do exactly that just because they (think they) can, so I agree. I will change the range of prefix_url now.

Hum, on second thought, not so sure about that.

The `prefix_url` values in the curie map are used to construct the IRIs in all slots typed as EntityReference. The way the `linkml:Uriorcurie` type (which is the underlying type of `EntityReference`) is defined, it has _always_ been possible to have “IRIs” that are in fact relative URI references and not (absolute) URIs.

So, something like this:

```
#curie_map:
#   PFX1: relative/uri/prefix
#   PFX2: another/relative/uri/prefix
#subject_id   predicate_id      object_id   mapping_justification
PFX1:1234     skos:exactMatch   PFX2:5678   semapv:manualMappingCuration
```

where the final identifiers are of the form `relative/uri/prefix/1234` or `another/relative/uri/prefix/5678` is in fact (and **has always been**) completely valid and accepted.

So I don’t think we should change that now. _We_ may think that semantic identifiers should always be fully resolvable IRIs, and that relative identifiers like in the example above are not great, but if people do have such identifiers they should be allowed to use them in SSSOM.

matentzn (approved):
Looks great, thank you!

PR #455, opened 2025-06-06
Change publication_date to be dcterms:issued

nichtich:
Publication date in dcterms is not `created` but `issued`. See for instance <https://www.w3.org/TR/vocab-dcat-3/#Property:resource_release_date>

### DIFF
--- a/src/sssom_schema/schema/sssom_schema.yaml
+++ b/src/sssom_schema/schema/sssom_schema.yaml
@@ -576,7 +576,7 @@ slots:
       propagated: true
   publication_date:
     description: The date the mapping was published. This is different from the date the mapping was asserted.
-    slot_uri: dcterms:created
+    slot_uri: dcterms:issued
     range: date
     examples:
       - value: 2021-01-01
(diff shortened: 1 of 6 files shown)


gouttegd (approved):
Looks like the correct thing to do, I think.

Strictly speaking it is a breaking change, but since it only affects the RDF serialisation which has never been officially specified, I am fine with it.

src/sssom_schema/schema/sssom_schema.yaml
```diff
@@ -576,7 +576,7 @@ slots:
       propagated: true
   publication_date:
     description: The date the mapping was published. This is different from the date the mapping was asserted.
-    slot_uri: dcterms:created
+    slot_uri: dcterms:issued
```
matentzn, line 579: The PR should _only_ include this change, nothing else. Refreshing the model files is done automatically, and in a versioned way.

I agree with the change though, just remove all the other files from the diff please :)

matentzn (changes requested):

Issue #457, opened 2025-06-06
Change mapping_date to be dcterms:created


nichtich:
The use of `pav:authoredOn` only makes sense if `pav:createdOn` is used as well to differentiate two types dates, in addition to the publication date. SSSOM only has one type of date so there is no need to not use plain old `dcterms:created`.

Requires #456.

gouttegd:
Good. I’ll add it to the [RDF spec PR](https://github.com/mapping-commons/sssom/pull/469), since the change is only really meaningful for RDF (well, and for JSON-LD but since the JSON-LD spec does not exist yet… :p ).

Issue #456, opened 2025-06-06
Change publication_date to be dcterms:issued


nichtich:
See PR #455 and discussion https://github.com/mapping-commons/sssom/discussions/450#discussioncomment-13383589

matentzn:
Yes this is the more appropriate property, agreed.

matentzn:
Fixed by https://github.com/mapping-commons/sssom/pull/460

PR #458, opened 2025-06-06
Assign a property to ALL enum values.

gouttegd:
Related to https://github.com/mapping-commons/sssom/discussions/454

This PR ensures that all enum values (_except_ for the `mapping_cardinality_enum`) throughout the LinkML model have a property assigned to them. This is important for RDF serialisation, so that all enum values are consistently rendered as resource identifiers, instead of having some values being rendered as resource identifiers and some values being rendered as string literals.

I normally agree with @nichtich ’s sentiment (expressed in #450) that we should as much as possible reuse existing properties from outside SSSOM (as we already do, for example, for most values of the `entity_type_enum`) and avoid creating SSSOM-specific properties in the `https://w3id.org/sssom/` namespace. However, for all the enum values concerned in this PR, I am not aware of pre-existing properties that would be suitable, so I did in fact create a bunch of SSSOM-specific properties. I am open to replacing some of them by more appropriate existing properties if someone can suggest some.

Also open to discussing the naming of those properties.

Of note, it would be nice if those properties could resolve to something meaningful to users. In particular, for the values in `sssom_version_enum`, I think most people would reasonably expect `https://w3id.org/sssom/version1.0` to resolve to a frozen version of the full specification as it was in the 1.0 release (and likewise for `https://w3id.org/sssom/version1.1` and any futur version). But this is an infrastructure issue that I consider out of scope for this PR.

### DIFF
--- a/src/sssom_schema/schema/sssom_schema.yaml
+++ b/src/sssom_schema/schema/sssom_schema.yaml
@@ -75,24 +75,12 @@ enums:
       Not: Negating the mapping predicate. The meaning of the triple becomes subject_id is not a predicate_id match to object_id.
   mapping_cardinality_enum:
     permissible_values:
-      "1:1":
-        meaning: sssom:Cardinality_1_1
-        description: One-to-one mapping
-      "1:n":
-        meaning: sssom:Cardinality_1_n
-        description: One-to-many mapping
-      "n:1":
-        meaning: sssom:Cardinality_n_1
-        description: Many-to-one mapping
-      "1:0":
-        meaning: sssom:Cardinality_1_0
-        description: One-to-none mapping
-      "0:1":
-        meaning: sssom:Cardinality_0_1
-        description: None-to-one mapping
-      "n:n":
-        meaning: sssom:Cardinality_n_n
-        description: Many-to-many mapping
+      "1:1": One-to-one mapping
+      "1:n": One-to-many mapping
+      "n:1": Many-to-one mapping
+      "1:0": One-to-none mapping
+      "0:1": None-to-one mapping
+      "n:n": Many-to-many mapping
 
 types:
  EntityReference:
--- a/src/sssom_schema/schema/sssom_schema.yaml
+++ b/src/sssom_schema/schema/sssom_schema.yaml
@@ -27,8 +27,12 @@ default_range: string
 enums:
   sssom_version_enum:
     permissible_values:
-      "1.0": SSSOM specification version 1.0
-      "1.1": SSSOM specification version 1.1
+      "1.0":
+        meaning: sssom:version1.0
+        description: SSSOM specification version 1.0
+      "1.1":
+        meaning: sssom:version1.1
+        description: SSSOM specification version 1.1
   entity_type_enum:
     permissible_values:
       owl class:
@@ -59,6 +63,7 @@ enums:
       rdf property:
         meaning: rdf:Property
       composed entity expression:
+        meaning: sssom:ComposedEntityExpression
         description: This value indicates that the entity ID does not represent a single entity, but a composite involving several individual entities. This value MUST NOT be used in the predicate_type slot. This specifications does not prescribe how an ID representing a composite entity should be interpreted; this is left at the discretion of applications.
         see_also:
         - https://github.com/mapping-commons/sssom/issues/402
@@ -66,15 +71,28 @@ enums:
         
   predicate_modifier_enum:
     permissible_values:
+      meaning: sssom:NegatedPredicate
       Not: Negating the mapping predicate. The meaning of the triple becomes subject_id is not a predicate_id match to object_id.
   mapping_cardinality_enum:
     permissible_values:
-      "1:1": One-to-one mapping
-      "1:n": One-to-many mapping
-      "n:1": Many-to-one mapping
-      "1:0": One-to-none mapping
-      "0:1": None-to-one mapping
-      "n:n": Many-to-many mapping
+      "1:1":
+        meaning: sssom:Cardinality_1_1
+        description: One-to-one mapping
+      "1:n":
+        meaning: sssom:Cardinality_1_n
+        description: One-to-many mapping
+      "n:1":
+        meaning: sssom:Cardinality_n_1
+        description: Many-to-one mapping
+      "1:0":
+        meaning: sssom:Cardinality_1_0
+        description: One-to-none mapping
+      "0:1":
+        meaning: sssom:Cardinality_0_1
+        description: None-to-one mapping
+      "n:n":
+        meaning: sssom:Cardinality_n_n
+        description: Many-to-many mapping
 
 types:
  EntityReference:
--- a/src/sssom_schema/schema/sssom_schema.yaml
+++ b/src/sssom_schema/schema/sssom_schema.yaml
@@ -71,7 +71,8 @@ enums:
         
   predicate_modifier_enum:
     permissible_values:
-      Not: Negating the mapping predicate. The meaning of the triple becomes subject_id is not a predicate_id match to object_id.
+      Not:
+        description: Negating the mapping predicate. The meaning of the triple becomes subject_id is not a predicate_id match to object_id.
         meaning: sssom:NegatedPredicate
   mapping_cardinality_enum:
     permissible_values:
--- a/src/sssom_schema/schema/sssom_schema.yaml
+++ b/src/sssom_schema/schema/sssom_schema.yaml
@@ -71,8 +71,8 @@ enums:
         
   predicate_modifier_enum:
     permissible_values:
-      meaning: sssom:NegatedPredicate
       Not: Negating the mapping predicate. The meaning of the triple becomes subject_id is not a predicate_id match to object_id.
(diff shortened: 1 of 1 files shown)


matentzn (dismissed):
Nothing controversial or breaking here?

src/sssom_schema/schema/sssom_schema.yaml
```diff
@@ -59,22 +63,36 @@ enums:
...
+        meaning: sssom:Cardinality_n_1
+        description: Many-to-one mapping
+      "1:0":
+        meaning: sssom:Cardinality_1_0
+        description: One-to-none mapping
+      "0:1":
+        meaning: sssom:Cardinality_0_1
+        description: None-to-one mapping
+      "n:n":
+        meaning: sssom:Cardinality_n_n
```
nichtich: Why not   `sssom:Cardinality_1:1`, `sssom:Cardinality_1:n`...  `sssom:Cardinality_n:n` so the URI can directly be derived from the enum value? Colon is not allowed in CURIE but allowed in URI.

Best existing URIs I could find are Wikidata items http://www.wikidata.org/entity/Q3424039 (1:1), http://www.wikidata.org/entity/Q45406573 (1:n), http://www.wikidata.org/entity/Q3058002 (n:m) but there is no 1:0 nor 0:1.

src/sssom_schema/schema/sssom_schema.yaml
```diff
@@ -27,8 +27,12 @@ default_range: string
...
     permissible_values:
-      "1.0": SSSOM specification version 1.0
-      "1.1": SSSOM specification version 1.1
+      "1.0":
+        meaning: sssom:version1.0
+        description: SSSOM specification version 1.0
+      "1.1":
+        meaning: sssom:version1.1
+        description: SSSOM specification version 1.1
   entity_type_enum:
```
nichtich, line 36: Version numbers are nothing special to SSSOM. See [this discussion](https://github.com/w3c/dxwg/issues/1280) that led to reusing `owl:versionInfo`.

src/sssom_schema/schema/sssom_schema.yaml
```diff
@@ -59,22 +63,36 @@ enums:
       rdf property:
         meaning: rdf:Property
       composed entity expression:
+        meaning: sssom:ComposedEntityExpression
```
nichtich, line 66: Composed entity expressions to model coordinated concepts are an ugly abnormity anyway, so I don't mind. In JSKOS there is no special type for composed entities but [concept bundles](https://gbv.github.io/jskos/#concept-bundles) are used for several purposes. Allowing slot `subject_id` and `object_id` to have multiple values might better solve this but nevermind.

nichtich (changes requested):
Encoding of version number should be changed to keep it a string value.

graybeal:
I don't like to bring this up, but please consider the following: is version really an enum?

For comparison, I assume we would agree that versionDate (or releaseDate or creationData) are not enums, because they could be anything, and everyone knows what a date format is. So we use a string for that.

But they couldn't be 'anything' in the real world of a particular system or entity. versionDate could only be one of a small set of strings that have been issued when the system is released. And every time you do a new version, you anoint another date as a versionDate, so the number of dates increases by 1. And you don't necessarily know ahead of time _which_ string will be the next date, but you know how to generate it. 

So what is the difference between versionDate and version, that makes versionDate a string and version an enum? 

I'd argue that in the same way that there can be any number of versionDates in the wild (beyond the ones we end up using for a given system) , there can be any number of versions, even semantic versions. 0.0.0 through infinite #.#.#. Is the idea that by giving version strings enums we have somehow made the world more interoperable? 

In the justification above, it seems we are using the version for a second purpose than just indicating the version: 

> I think most people would reasonably expect https://w3id.org/sssom/version1.0 to resolve to a frozen version of the full specification as it was in the 1.0 release

I agree with that statement on its own. But it's not because this is a _version_ identifier, it's because this is a versioned IRI for the _specification_. So `"https://w3id.org/sssom/version1.0"` should resolves as you say. For me it is only meaningful as a full IRI string; referring to it as "sssom:version1.0" would not work in a browser, would not be discoverable in a text search as an IRI, would not be expanded in most tools that are displaying the string—it only works in a context that the prefix has been defined and is being expanded. 

The point is that I've seen versionedIRIs and version IRIs and specification IRIs, but I've never seen any of them treated as an enum (unless it's to create a drop down menu to select your specification version, I suppose) and I've never seen any of them in the wild expressed as a CURIE. They are always strings.

Sorry to make this controversial. If you still think it's worth it the sky won't fall if you do it, I just have a different take on whether it will be helpful in the bigger scheme of things.

src/sssom_schema/schema/sssom_schema.yaml
gouttegd: Yes they are. They serve to identify a precise version of the spec, which I think is better done, in RDF context, with a IRI than with a string.

I would have thought that you would agree with that, given that you are the one who has been requesting that SSSOM-Py stop render some identifier slots as xsd:anyURI literals but as resource identifiers instead.

nichtich:
@gouttegd wrote:

> but the version of the SSSOM specification that the mapping set declares itself to be compliant with.

You are right, in this case the version could also be treated differently. Either an URI as suggested, or *not included in RDF at all* just like XML version number is not included in most DOM representations of the document tree. In the end, this issue seems of less relevance compared to how to express the actual content of a mapping set.

src/sssom_schema/schema/sssom_schema.yaml
gouttegd: After a discussion with @matentzn , we agreed that in fact there is no need for the values of the mapping cardinality enum to be represented as resource identifiers.

src/sssom_schema/schema/sssom_schema.yaml
matentzn: I like this. This allows me to _refer_ to a specific version of the spec in an RDF context, and I can use for example something like 

my:mapping_set dcterms:conformsTo sssom:version1.0 (outside of sssom)

src/sssom_schema/schema/sssom_schema.yaml
matentzn: I think this is needed for sure; 

> Allowing slot subject_id and object_id to have multiple values might better solve this but nevermind

This is not enough, but not a discussion for this issue here :P

src/sssom_schema/schema/sssom_schema.yaml
gouttegd: > (outside of sssom)

It could in fact be “inside of SSSOM“.

I was not aware of this `dcterms:conformsTo` property, but since it exists we could make it the property to use to render the `sssom_version` slot in RDF (the same way as, for example, the `subject_id` slot is rendered with `owl:annotatedSubject`).

All that would be needed is to amend the definition of the `sssom_version` slot as follows:

```diff
 slots:
   sssom_version:
     description: The version of the SSSOM specification a mapping set is compliant with.
     range: sssom_version_enum
+    slot_uri: dcterms:conformsTo
     instantiates: sssom:Versionable
     annotations:
       added_in: "1.1"
```

matentzn: (((((I did not know ```'''diff``` blocks, how awesome is that))))

> we could make it the property to use to render the sssom_version slot in RDF

I _think_ I would like that!

gouttegd: I’d suggest doing that at the same time we actually specify the RDF serialisation.

It would probably be a good idea at that time to review _all_ the properties that are currently associated with slots to ensure they are really the most appropriate one (as we have started to do in #456, #457).

matentzn: Sounds good!

src/sssom_schema/schema/sssom_schema.yaml
```diff
@@ -59,13 +63,15 @@ enums:
...
         description: This value indicates that the entity ID does not represent a single entity, but a composite involving several individual entities. This value MUST NOT be used in the predicate_type slot. This specifications does not prescribe how an ID representing a composite entity should be interpreted; this is left at the discretion of applications.
         see_also:
         - https://github.com/mapping-commons/sssom/issues/402
         - https://github.com/mapping-commons/sssom/blob/master/examples/schema/composite-entities.sssom.tsv
         
   predicate_modifier_enum:
     permissible_values:
+      meaning: sssom:NegatedPredicate
       Not: Negating the mapping predicate. The meaning of the triple becomes subject_id is not a predicate_id match to object_id.
```
ehartley:
```suggestion
      Not: Negating the mapping predicate. The meaning of the triple becomes subject_id is not a predicate_id match to object_id.
        meaning: sssom:NegatedPredicate
```
Moved the *meaning* to be after the enum value and indented one more level.

gouttegd: :scream: Thanks for spotting that!

matentzn (approved):

ehartley (approved):
Looks good!

PR #459, opened 2025-06-12
Add mapping_tool_id slot to model

### LINKED ISSUE #449: [New metadata element]: mapping_tool_id
**Element id:**

```
mapping_tool_id
```

**Value data type:**

```
URI
```

**Description**

URI of the mapping tool. Current slot `mapping_tool` can hold any string so it is not suitable for identification.

```
# license: https://w3id.org/sssom/license/unspecified
# mapping_set_id: https://example.org/id
# mapping_tool: Cocoda
# mapping_tool_id: https://coli-conc.gbv.de/cocoda/
# mapping_tool_version: 1.0.0
```

This would map to RDF more or less like

~~~ttl
<https://example.org/id> sssom:mapping_tool <https://coli-conc.gbv.de/cocoda/> .
<https://coli-conc.gbv.de/cocoda/> dct:title "Cocoda" ; owl:versionInfo "1.0.0" .
~~~

matentzn:
Resolves #449 

- [X] tests have been added/updated (if applicable)

- [X] make sure any new slot is annotated with the appropriate `added_in` annotation

This PR is adding mapping_tool_id as an entity reference to the SSSOM model. I have also included a note that if possible, this slot should be preferred over the less formal `mapping_tool` slot.

cc @nichtich @gouttegd

### DIFF
--- a/examples/schema/mapping_tool_id.sssom.tsv
+++ b/examples/schema/mapping_tool_id.sssom.tsv
@@ -1,0 +1,14 @@
+#curie_map:
+#  HP: http://purl.obolibrary.org/obo/HP_
+#  MP: http://purl.obolibrary.org/obo/MP_
+#  orcid: https://orcid.org/
+#  wikidata: https://www.wikidata.org/wiki/
+#mapping_set_id: https://w3id.org/sssom/commons/examples/mapping_tool_id.sssom.tsv
+#license: "https://creativecommons.org/publicdomain/zero/1.0/"
+#creator_id: orcid:0000-0002-7356-1779
+#mapping_provider: "https://w3id.org/sssom/core_team"
+#issue_tracker: "https://github.com/mapping-commons/mh_mapping_initiative/issues"
+#comment: This is an example file for the SSSOM for illustration only. Its contents are entirely fabricated.
+subject_id	predicate_id	object_id	mapping_justification	mapping_tool_id
+HP:0009124	skos:exactMatch	MP:0000003	semapv:ManualMappingCuration	wikidata:Q58057366
+HP:0008551	skos:exactMatch	MP:0000018	semapv:ManualMappingCuration	wikidata:Q58057366
--- a/examples/schema/mapping_tool_id.sssom.tsv
+++ b/examples/schema/mapping_tool_id.sssom.tsv
@@ -5,7 +5,8 @@
 #  wikidata: https://www.wikidata.org/wiki/
 #mapping_set_id: https://w3id.org/sssom/commons/examples/mapping_tool_id.sssom.tsv
 #license: "https://creativecommons.org/publicdomain/zero/1.0/"
-#creator_id: orcid:0000-0002-7356-1779
+#creator_id:
+#  - orcid:0000-0002-7356-1779
 #mapping_provider: "https://w3id.org/sssom/core_team"
 #issue_tracker: "https://github.com/mapping-commons/mh_mapping_initiative/issues"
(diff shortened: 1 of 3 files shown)


examples/schema/mapping_tool_id.sssom.tsv
```diff
@@ -0,0 +1,14 @@
+#curie_map:
+#  HP: http://purl.obolibrary.org/obo/HP_
+#  MP: http://purl.obolibrary.org/obo/MP_
+#  orcid: https://orcid.org/
+#  wikidata: https://www.wikidata.org/wiki/
+#mapping_set_id: https://w3id.org/sssom/commons/examples/mapping_tool_id.sssom.tsv
+#license: "https://creativecommons.org/publicdomain/zero/1.0/"
+#creator_id: orcid:0000-0002-7356-1779
```
gouttegd:
```suggestion
#creator_id:
#  - orcid:0000-0002-7356-1779
```

It’s unrelated to the issue this PR is addressing, but please avoid using `creator_id` as if it were a single-valued slot. Yes, we have explicitly [allowed scalar values](https://mapping-commons.github.io/sssom/spec-formats-tsv/#multi-valued-slots-with-a-single-value) for multi-valued slots, but we only did so for compatibility with existing sets. Our own examples should do the right thing.

src/sssom_schema/schema/sssom_schema.yaml
```diff
@@ -548,6 +548,7 @@ slots:
   mapping_tool:
     description: A reference to the tool or algorithm that was used to generate the
       mapping. Should be a URL pointing to more info about it, but can be free text.
+      Consider using the mapping_tool_id slot for a more standardised reference.
```
gouttegd, line 551: Nitpick: I’d personally use a stronger word than “consider” here. If a semantic identifier is available for the tool, then `mapping_tool_id` SHOULD be used instead.

src/sssom_schema/schema/sssom_schema.yaml
```diff
@@ -557,6 +558,20 @@ slots:
         description: A URL pointing to the AgreementMakerLight project.
       - value: AgreementMakerLight
         description: A string (name) denoting the AgreementMakerLight project.
+  mapping_tool_id:
+    description: The ID (entity reference) of the tool or algorithm that was used to generate the
+      mapping.
+    range: EntityReference
+    instantiates: sssom:Propagatable
```
gouttegd:
```suggestion
    instantiates:
      - sssom:Propagatable
      - sssom:Versionable
```

You need to instantiate `sssom:Versionable` as well in order to use the `added_in` annotation.

gouttegd (changes requested):

src/sssom_schema/schema/sssom_schema.yaml
gouttegd: (Well, actually for now you do not _need_ it, since as far as I know LinkML currently does not perform any kind of checks on the annotations – in particular, it doesn’t check that the annotations effectively used match those defined by the instantiated extensions. But that may change with future LinkML versions.)

gouttegd (approved):

nichtich (dismissed):
Looks good. Nitpicking comments in the issue #449.

matentzn:
@nichtich I added you to the SSSOM core team so your reviews can count; 

For the rest, I will merge this PR as is anytime after Wednesday 25th.

gouttegd (approved):

PR #460, opened 2025-06-14
Map sssom:publication_date to dcterms:issued

### LINKED ISSUE #456: Change publication_date to be dcterms:issued

matentzn:
dcterms:issues is the more appropriate property to use, which is defined as "Date of formal issuance of the resource."

https://www.dublincore.org/specifications/dublin-core/dcmi-terms/#http://purl.org/dc/terms/issued

Resolves #456 

- [ ] ~~`make test` has been run locally~~

@gouttegd I don't think we need to add verioning annotations when we make a change like this, right?


gouttegd (approved):

gouttegd:
> I don't think we need to add versioning annotations when we make a change like this, right?

Correct.

But when the RDF serialisation will have been formally specified, then this kind of change will still require a notice in the [backwards compatibility notes](https://mapping-commons.github.io/sssom/spec-model/#model-changes-across-versions). This is not needed here only because we don’t have a spec for RDF yet.

nichtich (approved):

Issue #406, opened 2024-12-23
Contradiction in license Element Description for MappingSet Class


ruihuang19:
In the description of the license element on https://mapping-commons.github.io/sssom/license/, it is stated as an optional element. However, in the "Induced" section at the bottom of https://mapping-commons.github.io/sssom/MappingSet/, the required property of the license element is set to true.

![image](https://github.com/user-attachments/assets/61890454-5e17-4bd1-9f72-50ed02bd4ff0)
![image](https://github.com/user-attachments/assets/bc3426e1-7838-4a44-b6cb-a0ed81e3f6e8)
![image](https://github.com/user-attachments/assets/9edffdfe-9850-45d7-9206-70b0a259d5fb)

matentzn:
@ruihuang19 thank you for the issue!

@sujaypatil96 I will need your help for this.

In my template I use

```
{% for c in gen.all_class_objects()|sort(attribute=sort_by) -%}
{% if c.name == "mapping set" %}
{{gen.link(c)}}: {{c.description|enshorten}}

| Column/Field       | Description                                             | Required    |
|--------------------|---------------------------------------------------------|-------------|
{%- for slot in c.slots %}
{%- set slot_info = schemaview.get_slot(slot) %}
| **{{ gen.link(slot) }}**     | {{ slot_info.description | default("No description") }} | {% if slot_info.required | default(false) %}Required{% elif slot_info.recommended | default(false) %}Recommended{% else %}Optional{% endif %} |
{%- endfor %}
{%- endif %}
{%- endfor %}
```

To show a row in the overall table. `schemaview.get_slot(slot)` retrieves the _general_ information about the slot, but I want to know the information about the slot in the context for `c`. So `license` is not _required_ in general, but in the context of the `mapping set` class, it is.

sujaypatil96:
I think the documentation pages are showing the correct behavior here?

The license slot page: https://mapping-commons.github.io/sssom/license/ is "optional" (_required_ has not been set to _true_) which is a correct reflection of the linkml spec. for that slot as we can see here: https://mapping-commons.github.io/sssom/license/#schema-developer-documentation

However, it is being modified to `required: true` using `slot_usage` in the context of the `MappingSet` class. See the spec. here: https://mapping-commons.github.io/sssom/MappingSet/#induced

[...]

PR #462, opened 2025-06-16
Upgrade GitHub Actions workflows

sujaypatil96:
This PR upgrades the the GitHub Actions workflows in the repo to use the latest versions of various Actions, poetry configuration steps, documentation build and deployment, etc.

Required to address #406 

[Add a description, mentioning at least relevant #ISSUE and how it was addressed. A bulleted list of all changes performed by the PR is is helpful.]


.github/workflows/deploy_documentation.yml
```diff
@@ -4,7 +4,11 @@ name: Deploy Documentation
 on:
   workflow_dispatch:
   push:
-    branches: [ master ]
+    branches:
+      - master
+  pull_request:
```
matentzn: i dont really want to deploy the documentation on pull request; that's a weird pattern.

.github/workflows/pypi-publish.yaml
```diff
@@ -11,26 +11,36 @@ jobs:
...
 
+    #----------------------------------------------
+    #             install dependencies
+    #----------------------------------------------
     - name: Install dependencies
       run: poetry install --no-interaction
 
     - name: Build source and wheel archives
-      run: |
-        poetry version $(git describe --tags --abbrev=0)
```
matentzn, line 32: So where does it get the version form now?

matentzn, line 32: So where does it get the version form now?

.github/workflows/deploy_documentation.yml
sujaypatil96: Makes sense. We have a workflow in our repos, for example: https://github.com/microbiomedata/nmdc-schema/blob/main/.github/workflows/test_pages_build.yaml which allows you to preview documentation changes upon making a PR (Action creates a public URL and adds it as a comment on the PR). Let me know if you want that added as a feature to this repo?

For now I'll remove documentation deployment upon PR's. It would be useful if you were adding something like an FAQ, updating the contributor guidelines, policy updates, etc. and you wanted to see it on the website prior to a release, but if that's not a use case at the moment, no worries.

.github/workflows/pypi-publish.yaml
sujaypatil96: It's currently picking up on the version number based on the tag name, resetting the `version` number here: https://github.com/mapping-commons/sssom/blob/master/pyproject.toml#L3 and using that `version` number to build the package.

sujaypatil96: It's currently picking up on the version number based on the tag name, resetting the version number here: https://github.com/mapping-commons/sssom/blob/master/pyproject.toml#L3 and using that version number to build the package.

sujaypatil96:
@matentzn this PR is ready for a second round of review.

.github/workflows/pypi-publish.yaml
matentzn: So `poetry version $(git describe --tags --abbrev=0)` was simply redundant? Or was this somehow made possible through the above

```
 pipx inject poetry poetry-dynamic-versioning
```

matentzn (approved):
This looks good, I still have my one comment which I would just like to have answered for my own understanding but I am sure its fine. I will get a second reviewer.

.github/workflows/pypi-publish.yaml
dalito: The latter. `poetry version` is not required anymore because `poetry-dynamic-versioning` derives the version from the latest tag.

dalito:
Adding an environment name in the trusted publishing setup is strongly encouraged. In the case here it should be `pypi-release`. The other changes looked fine to me.

Not related to the changes but it is surprising to see continued support for unsupported Python 3.8 and no support for any version newer than 3.10.

.github/workflows/pypi-publish.yaml
sujaypatil96: Yup, that's right, thanks for clarifying that @dalito 😁

matentzn: thank you both!

Issue #421, opened 2025-03-21
RDF serialisation needs a specification


gouttegd:
If the RDF format produced/consumed by `sssom-py` is supposed to be usable at the same level as the SSSOM/TSV format, it needs to be specified.

Having to infer the serialisation rules by looking at the output of the “reference implementation“ (which is what I had to do to implement RDF support in SSSOM-Java) is not acceptable for something that pretends to be a standard.

Same thing for the OWL/RDF serialisation, which is only [specified (poorly)](https://mapping-commons.github.io/sssom/spec-formats-owl/) for the individual mappings, not for mapping set objects.

matentzn:
I think this is critical for some of our prospective user base. for me, the main problem is how to deal with #359 when the pressure mounts (I expect it to mount the latest in the summer). will we have "mapping_id" (as an "optional" property) hang off the reified RDF statement as sssom:mapping_id, or will we use it as the resource identifier for the mapping (which is more correct, but will create an entirely different RDF serialisation). 

I expect people to favour the latter, but I find it hard to reconcile this with the "optionality" of the mapping_id.

gouttegd:
> we use [the `mapping_id`] as the resource identifier for the mapping (which is more correct, but will create an entirely different RDF serialisation). I expect people to favour the latter

As you note, this would result in a RDF serialisation entirely different _than the one **currently** being supported by SSSOM-Py_ (and SSSOM-Java, since in the absence of a spec SSSOM-Java basically did the same thing as SSSOM-Py).

That is not necessarily a blocker (RDF serialisation is _not_ specified as of SSSOM 1.0, so it’s fine to radically change it), but still something to be carefully considered.

If people have already started using the current RDF serialisation, we might end up having to specify _two_ different RDF serialisations: one that corresponds to the current output produced by SSSOM-Py (where individual mappings are blank nodes, without a resource identifier), and one that uses the to-be-introduced `mapping_id` slot as the resource identifier.

> I find it hard to reconcile this with the "optionality" of the mapping_id.

(A) The new RDF serialisation format is only available for mapping sets that do use the new `mapping_id` slot. The slot itself remains generally optional, but the new RDF serialisation format requires it.

(B) If you have a mapping set without `mapping_id` slots and you absolutely want to serialize it into the new RDF format, you must generate `mapping_id` slots on the fly.

gouttegd:
> I find it hard to reconcile this with the "optionality" of the mapping_id.

[...]

Issue #399, opened 2024-11-04
Change `license` element to support CURIEs, e.g., for SPDX


cthoyt:
The license element of the schema is currently a URI. It would be nice to be able to specify licenses based on SPDX identifiers as well, since this has the benefit of being a controlled vocabulary, and pushing people towards using well-defined licenses

gouttegd:
In principle no objection to that, but it should be noted that, with the current implementation in SSSOM-Py, the “as well” bit is not possible: either the slot expects a URI, _or_ it accepts a CURIE, but it cannot accept both: SSSOM-Py _enforces_ any slot typed as a `EntityReference` MUST be in CURIE form.

So if we change the type of `license` to be an `EntityReference`, SSSOM-Py will no longer accept the use of plain IRI, and will barf on a mapping set with

```
#license: https://creativecommons.org/licenses/by/4.0
```

(which is currently perfectly valid). This will have to be replaced by a pseudo-CURIE, as in

```
#curie_map:
#  CCORG: https://creativecommons.org/licenses/
#license: CCORG:by/4.0
```

I’m very reluctant to break existing mapping sets just for that, so I will oppose any such change until/unless SSSOM-Py is made to silently accept the use of IRIs for EntityReference-typed slots.

matentzn:
Has anyone ever tried to validate

```
#curie_map:
#  CC.ORG: "CC.ORG:"
#license: CC.ORG:by/4.0
```

Just out of curiosity for a hacky workaround for some cases. In general if we type something "uriorcurie" its is my understanding it should literally be possible to either or, even if tools don't support it yet.

All that said,

is this really that bad?
 
```
#curie_map:
#  SPDX: https://spdx.org/licenses/
#license: SPDX:AdaCore-doc
```

TBH SPDX may not be _as_ strong a case for allowing curies explicitly, even if the RDF transformation of the above adds https://spdx.org/licenses/AdaCore-doc as the license..

gouttegd:
> Has anyone ever tried to validate
>
> ```
> #curie_map:
> #  CC.ORG: "CC.ORG:"
> #license: CC.ORG:by/4.0
```

Err, yes, this works, but why would ever do that?

If you really want to store the license as a CURIE, you can do so the “normal” way, without having to make this weird “prefix” declaration:

```
#curie_map:
#  CCORG: https://creativecommons.org/licenses
#license: CCORG/by/4.0
```

[...]

PR #469, opened 2025-07-14
Initial draft of the SSSOM/RDF spec.

gouttegd:
Resolves [#421]

This is the complete proposal for the specification of the SSSOM/RDF serialisation format, according to the current state of the discussions about it.

### DIFF
--- a/src/docs/spec-formats-rdf.md
+++ b/src/docs/spec-formats-rdf.md
@@ -1,0 +1,366 @@
+# The SSSOM/RDF serialisation format
+
+This section defines how to represent a SSSOM mapping set as a [RDF
+model](https://www.w3.org/TR/rdf11-concepts/).
+
+
+## RDF formats
+The RDF model that represents a SSSOM mapping set is independent of the
+concrete format that may be used to serialise the model.
+
+It is RECOMMENDED that implementations support reading and writing a
+SSSOM set from and to the [RDF Turtle](https://www.w3.org/TR/turtle/)
+format at least. They MAY support any other RDF concrete format (e.g.
+RDF/XML, TriG, N-Triples, etc.).
+
+This specification does not mandate how a concrete RDF syntax is to be
+used. For example, if the RDF syntax allows named resources and
+predicates to be serialised as either IRIs or CURIEs, if is left at the
+discretion of the implementations (or their users) to decide which form
+to use.
+
+<a id="sssom-slots"></a>
+## Representation of slots
+A metadata slot on any given SSSOM object (such as a `Mapping` or a
+`MappingSet`) is represented as a RDF triple where:
+
+* the subject is the resource representing the SSSOM object;
+* the predicate is either:
+    * the property indicated by the `URI` field in the LinkML
+      description of the slot, if such a field is present;
+    * or a property constructed by catenating the
+      `https://w3id.org/sssom/` namespace and the name of the slot;
+* the object is the value of the slot.
+
+### Representation of slot values
+The following rules determine how the value of a slot is represented as
+the object of a RDF triple.
+
+#### For slots typed as `sssom:EntityReference`
+(e.g. `subject_id`, `mapping_justification`, `subject_source`…)
+
+The value is rendered as a named RDF resource (IRI).
+
+#### For slots typed as `sssom:NonRelativeURI`
+(e.g. `license`, `mapping_provider`, `issue_tracker`…)
+
+The value is rendered as a named RDF resource (IRI).
+
+#### For slots typed as `linkml:date`
+(e.g. `mapping_date`, `publication_date`)
+
+The value is represented as a `xsd:date` literal.
+
+#### For slots typed as `linkml:double`
+(e.g. `mapping_set_confidence`, `confidence`, `similarity_score`)
+
+The value is represented as a `xsd:double` literal.
+
+#### For slots typed as an enumeration
+(e.g. `sssom_version`, `mapping_cardinality`, `subject_type`…)
+
+If the permissible values for the enumeration are defined in the LinkML
+model as having an associated `meaning` property, then the value is
+represented as a named RDF resource with the indicated property.
+Otherwise, the value is represented as a string literal.
+
+#### For slots typed as a SSSOM object
+(e.g. `mappings`, `extension_definitions`)
+
+The value is represented as a RDF resource. Whether the resource is
+named (IRI) or not (blank node) will depend on the type of the object,
+see the [section on representing SSSOM objects](#sssom-objects) below
+for details.
+
+### Representation of multi-valued slots
+(e.g. `creator_id`, `see_also`, `object_match_field`…)
+
+As an exception to the general principle that slots are represented by a
+single RDF triple, multi-valued slots are represented by as many
+triples as there are values, each value being the object of one triple.
+
+> Non-normative notes:
+>
+> 1. This means, in particular, that RDF complex structures intended to
+>    represent collections of values, such as `rdfs:Container` or
+>    `rdfs:List`, MUST NOT be used to represent multi-valued SSSOM
+>    slots.
+> 2. This also implies that values in multi-valued slots are _not_
+>    ordered.
+
+The other rules above apply to determine how each single value is to be
+represented.
+
+<a id="extension-slots"></a>
+### Representation of extension slots
+An [extension slot](spec-model.md#non-standard-slots) is represented in
+a similar way to a standard slot, with the following specific rules.
+
+The predicate is the property associated to the extension slot, as
+indicated by the `property` slot in the set’s
+[definition](ExtensionDefinition.md) of the extension.
+
+The value of the extension is represented:
+
+* as a named RDF resource, if the `type_hint` of the extension
+  definition is `linkml:uriOrCurie`;
+* otherwise, as a literal of the type indicated by the `type_hint`.
+
+
+<a id="sssom-objects"></a>
+## Representation of SSSOM objects
+
+### Representation of a `Mapping` object
+The RDF type of a `Mapping` object is `owl:Axiom`.
+
+If the `Mapping` object has a `record_id` slot, then the value of that
+slot is used as the named RDF resource that represents the object (and
+consequently, that slot MUST NOT be represented using the [general
+rules](#sssom-slots) for the representation of slots as defined above).
+Otherwise, the `Mapping` object is represented as a blank node.
+
+### Representation of a `MappingSet` object
+The RDF type of a `MappingSet` object is `sssom:MappingSet`.
+
+A `MappingSet` object is represented by a named RDF resource
+corresponding to the value of the `mapping_set_id` slot (which
+consequently MUST NOT be represented using the [general
+rules](#sssom-slots) for the representation slots as defined above).
+
+The `curie_map` slot MUST NOT be represented using the [general
+rules](#sssom-slots). Instead, if it is needed it MUST be represented
+using whatever mechanism is provided by the concrete RDF serialisation
+format (e.g. `@prefix` declarations in [RDF
+Turtle](https://www.w3.org/TR/turtle/) or [RDF
+TriG](https://www.w3.org/TR/trig/), or `xmlns` namespace declarations in
+[RDF/XML](https://www.w3.org/TR/rdf-syntax-grammar/)).
+
+> Non-normative notes
+>
+> 1. The CURIE map may not be needed at all if all named resources and
+>    predicates are always serialised as full-length IRIs.
+> 2. If at least some named resources or predicates are serialised as
+>    CURIEs, the RDF requirement that all used prefix names must be
+>    declared (using the appropriate mechanism for the chosen concrete
+>    syntax) takes precedence over the possibility of omitting the
+>    declarations of prefix names that are considered
+>    [built-in](spec-intro.md#iri-prefixes) in the context of SSSOM.
+
+### Representation of a `ExtensionDefinition` object
+The RDF type of a `ExtensionDefinition` object is
+`sssom:ExtensionDefinition`.
+
+A `ExtensionDefinition` object has no identifier of any kind and is
+always represented by a blank node.
+
+## Special considerations for serialising to RDF
+When serialising a mapping set to SSSOM/RDF, implementations should
+consider how the resulting RDF file is intended to be used. In
+particular, they should ponder whether it is expected that the RDF
+serialisation can at any time be converted back to any other SSSOM
+format (e.g. SSSOM/TSV), or if it is only intended to be used by
+“generic”, non-SSSOM-aware RDF applications.
+
+Depending on that intended usage (if it is known), implementations may
+adopt slightly different behaviours as described in the following
+subsections.
+
+### Serialisations of identifiers
+If the serialisation is intended to be convertible back to another SSSOM
+format (especially the SSSOM/TSV format), implementations MUST
+serialise identifiers as CURIEs and include the required prefix
+declarations.
+
+> Non-normative explanation
+>
+> This is because, if all identifiers are serialised as full-length
+> IRIs, then even if the RDF file includes prefix declarations, they may
+> be stripped away by a RDF reader, since they are not needed. And
+> without those prefix declarations, it would not be possible to
+> serialise the set back as a SSSOM/TSV file (remember that the
+> SSSOM/TSV format _requires_ that identifiers be serialised as CURIEs).
+
+Conversely, if the ability to convert the RDF file back to another SSSOM
+format is not required, implementations can freely decide whether to
+serialise identifiers as IRIs or CURIEs (assuming the concrete RDF
+syntax allows that of course).
+
+### Extension definitions
+Extension definitions MAY be omitted if the RDF file is only intended to
+be used by RDF applications.
+
+Conversely, they SHOULD be included if the set is intended to be
+convertible back to another SSSOM format.
+
+> Non-normative explanation
+>
+> The whole point of an extension definition in SSSOM is to provide (1)
+> a property that confers some meaning to the extension, and (2) the
+> type of the expected values. In RDF, as described
+> [above](#extension-slots), those two bits of information are already
+> contained in the triple that represent the extension slot, so there is
+> no need for an additional definition.
+>
+> But the extension definition also provides the `slot_name` which is
+> used to represent the extension slot in other formats (especially
+> SSSOM/TSV), so if conversion back to other SSSOM formats is required,
+> ensuring that the extension definitions are present in the RDF
+> serialisation is helpful.
+
+### Propagation and condensation
+Propagatable slots can be represented in RDF indifferently in their
+propagated or condensed form, following the [normal
+rules](spec-model.md##propagation-of-mapping-set-slots) for propagation
+and condensation.
+
+But if the RDF file is intended to be used by generic, non-SSSOM-aware
+RDF applications, then implementations SHOULD serialise propagatable
+slots in their propagated form.
+
+> Non-normative explanation
+>
+> Propagation is a SSSOM-specific concept. If a RDF application is
+> provided with a RDF file representing a set with condensed slots, the
+> application will not know to propagate the condensed slots at the set
+> level down to the level of the individual mappings, which will result
+> in the application having an incomplete view of the mappings.
+
+
+## Compatibility with pre-standard RDF representations
+The present specification of the SSSOM/RDF format differs slightly from
+what several implementations of SSSOM have been producing before the
+format was formally specified.
+
+In the name of backward compatibility, implementations MAY support the
+alternative rules described in the following subsections when
+deserialising from RDF.
+
+Implementations MUST NOT follow these rules when serialising to RDF.
+
+### Representation of slots typed as `sssom:NonRelativeURI`
+Implementations MAY accept a value represented as a `xsd:anyURI`
+literal.
+
+### Representation of slots typed as an enumeration
+Implementations MAY accept a value represented as a string literal, even
+if the value is defined in the LinkML model as having an associated
+`meaning` property.
+
+For example, implementations MAY accept
+
+```ttl
+?mapping sssom:predicate_modifier "Not"^^xsd:string .
+```
+
+as an alternative to
+
+```ttl
+?mapping sssom:predicate_modifier sssom:NegatedPredicate .
+```
+
+### Representation of a `MappingSet` object
+Implementations MAY accept a `MappingSet` object represented as a blank
+node, with the `mapping_set_id` slot being represented as any other
+slot.
+
+For example, instead of
+
+```ttl
+<https://example.org/myset> a sssom:MappingSet .
+```
+
+implementations MAY accept
+
+```ttl
+[] a sssom:MappingSet ;
+   sssom:mapping_set_id <https://example.org/myset> .
+```
+
+or even (by also applying the alternative rule regarding the
+representation of slots typed as `sssom:NonRelativeURI`)
+
+```ttl
+[] a sssom:MappingSet ;
+   sssom:mapping_set_id "https://example.org/myset"^^xsd:anyURI .
+```
+
+## Examples
+
+> This section is non-normative.
+
+Considering the following set in the SSSOM/TSV format:
+
+```
+#curie_map:
+#  EXT: https://example.org/properties/
+#  FOODON: http://purl.obolibrary.org/obo/FOODON_
+#  KF_FOOD: https://kewl-foodie.inc/food/
+#  ORCID: https://orcid.org/
+#mapping_set_id: https://example.org/sample-set
+#mapping_set_description: Manually curated alignment of KEWL FOODIE INC internal food and nutrition database with Food Ontology (FOODON). Intended to be used for ontological analysis and grouping of KEWL FOODIE INC related data.
+#license: https://creativecommons.org/licenses/by/4.0/
+#mapping_date: 2025-07-14
+#extension_definitions:
+#  - slot_name: ext_fooable
+#    property: EXT:isFooable
+#    type_hint: xsd:boolean
+subject_id	subject_label	predicate_id	object_id	object_label	mapping_justification	author_id	confidence	ext_fooable
+KF_FOOD:F001	apple	skos:exactMatch	FOODON:00002473	apple (whole)	semapv:ManualMappingCuration	ORCID:0000-0002-7356-1779	0.95	true
+KF_FOOD:F002	gala	skos:exactMatch	FOODON:00003348	Gala apple (whole)	semapv:ManualMappingCuration	ORCID:0000-0002-7356-1779	1	false
+```
+
+A valid serialisation of that set in RDF/Turtle would be:
+
+```ttl
+@prefix EXT: <https://example.org/properties/> .
+@prefix FOODON: <http://purl.obolibrary.org/obo/FOODON_> .
+@prefix KF_FOOD: <https://kewl-foodie.inc/food/> .
+@prefix ORCID: <https://orcid.org/> .
+@prefix dcterms: <http://purl.org/dc/terms/> .
+@prefix owl: <http://www.w3.org/2002/07/owl#> .
+@prefix pav: <http://purl.org/pav/> .
+@prefix semapv: <https://w3id.org/semapv/vocab/> .
+@prefix skos: <http://www.w3.org/2004/02/skos/core#> .
+@prefix sssom: <https://w3id.org/sssom/> .
+@prefix xsd: <http://www.w3.org/2001/XMLSchema#> .
+
+<https://example.org/sample-set> a sssom:MappingSet;
+  dcterms:description "Manually curated alignment of KEWL FOODIE INC internal food and nutrition database with Food Ontology (FOODON). Intended to be used for ontological analysis and grouping of KEWL FOODIE INC related data.";
+  dcterms:license <https://creativecommons.org/licenses/by/4.0/>;
+  sssom:extension_definitions [
+      sssom:property EXT:isFooable;
+      sssom:slot_name "ext_fooable";
+      sssom:type_hint xsd:boolean
+    ];
+  sssom:mappings [ a owl:Axiom;
+      pav:authoredBy ORCID:0000-0002-7356-1779;
+      pav:authoredOn "2025-07-14"^^xsd:date;
+      owl:annotatedProperty skos:exactMatch;
+      owl:annotatedSource KF_FOOD:F001;
+      owl:annotatedTarget FOODON:00002473;
+      EXT:isFooable true;
+      sssom:confidence 9.5E-1;
+      sssom:mapping_justification semapv:ManualMappingCuration;
+      sssom:object_label "apple (whole)";
+      sssom:subject_label "apple"
+    ], [ a owl:Axiom;
+      pav:authoredBy ORCID:0000-0002-7356-1779;
+      pav:authoredOn "2025-07-14"^^xsd:date;
+      owl:annotatedProperty skos:exactMatch;
+      owl:annotatedSource KF_FOOD:F002;
+      owl:annotatedTarget FOODON:00003348;
+      EXT:isFooable false;
+      sssom:confidence 1.0E0;
+      sssom:mapping_justification semapv:ManualMappingCuration;
+      sssom:object_label "Gala apple (whole)";
+      sssom:subject_label "gala"
+    ] .
+```
+
+Note that the two `Mapping` objects are represented as blank nodes,
+since the original set does not contain any `record_id` slot.
+
+Note also that (1) identifiers are serialised as CURIEs whenever
+possible, and (2) the definition for the `EXT:isFooable` extension is
+included. This means that the set can be fully converted back to
+SSSOM/TSV without any loss of information.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 - Add `mapping_tool_id` slot to the `Mapping` and `MappingSet` classes ([issue](https://github.com/mapping-commons/sssom/issues/449)).
 - Add `record_id` slot to the `Mapping` class ([issue](https://github.com/mapping-commons/sssom/issues/359)).
 - Change all URI-typed slots to clarify that they expect _non-relative_ URIs as values ([issue](https://github.com/mapping-commons/sssom/issues/448)).
+- Add specification for the RDF serialisation ([discussion](https://github.com/mapping-commons/sssom/discussions/454)).
 
 ## SSSOM version 1.0.0
 
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -39,8 +39,9 @@ nav:
     - Serialisations:
       - Introduction: spec-formats.md
       - SSSOM/TSV serialisation: spec-formats-tsv.md
+      - SSSOM/JSON serialisation: spec-formats-json.md
+      - SSSOM/RDF serialisation: spec-formats-rdf.md
       - OWL/RDF serialisation: spec-formats-owl.md
-      - JSON serialisation: spec-formats-json.md
   - Resources for contributors: contributing.md
   - Resources for users:
     - FAQ: faq.md
--- a/src/docs/spec-formats.md
+++ b/src/docs/spec-formats.md
@@ -3,7 +3,8 @@
 The SSSOM standard defines the following serialisation formats for storing and exchanging mapping sets:
 
 * the [SSSOM/TSV](spec-formats-tsv.md) format;
-* the [SSSOM JSON](spec-formats-json.md) format;
+* the [SSSOM/JSON](spec-formats-json.md) format;
+* the [SSSOM/RDF](spec-formats-rdf.md) format;
 * and the [OWL/RDF](spec-formats-owl.md) format.
 
 Implementations MUST support the SSSOM/TSV format. They MAY support the other formats.


src/docs/spec-formats-rdf.md
```diff
@@ -0,0 +1,366 @@
...
+<a id="sssom-slots"></a>
+## Representation of slots
+A metadata slot on any given SSSOM object (such as a `Mapping` or a
+`MappingSet`) is represented as a RDF triple where:
+
+* the subject is the resource representing the SSSOM object;
+* the predicate is either:
+    * the property indicated by the `URI` field in the LinkML
+      description of the slot, if such a field is present;
+    * or a property constructed by catenating the
```
matentzn:
```suggestion
    * or a property constructed by concatenating the
```

src/docs/spec-formats-rdf.md
```diff
@@ -0,0 +1,366 @@
...
+
+#### For slots typed as `sssom:EntityReference`
+(e.g. `subject_id`, `mapping_justification`, `subject_source`…)
+
+The value is rendered as a named RDF resource (IRI).
+
+#### For slots typed as `sssom:NonRelativeURI`
+(e.g. `license`, `mapping_provider`, `issue_tracker`…)
+
+The value is rendered as a named RDF resource (IRI).
```
matentzn: why use "is" instead of `MUST BE` lingo here?

src/docs/spec-formats-rdf.md
```diff
@@ -0,0 +1,366 @@
...
+named (IRI) or not (blank node) will depend on the type of the object,
+see the [section on representing SSSOM objects](#sssom-objects) below
+for details.
+
+### Representation of multi-valued slots
+(e.g. `creator_id`, `see_also`, `object_match_field`…)
+
+As an exception to the general principle that slots are represented by a
+single RDF triple, multi-valued slots are represented by as many
+triples as there are values, each value being the object of one triple.
```
matentzn: i stumbled across this sentence multiple times. Maybe this can be written clearer like:

for each value {v1,v..,vn} represented by a set of individual triples `{a,b,v1; a,b,v2,...a,b,vn}`.

src/docs/spec-formats-rdf.md
```diff
@@ -0,0 +1,366 @@
...
+
+The predicate is the property associated to the extension slot, as
+indicated by the `property` slot in the set’s
+[definition](ExtensionDefinition.md) of the extension.
+
+The value of the extension is represented:
+
+* as a named RDF resource, if the `type_hint` of the extension
+  definition is `linkml:uriOrCurie`;
+* otherwise, as a literal of the type indicated by the `type_hint`.
```
matentzn: what if the literal is xsd:anyUri or `sssom:NonRelativeURI`?

src/docs/spec-formats-rdf.md
```diff
@@ -0,0 +1,366 @@
...
+<a id="sssom-objects"></a>
+## Representation of SSSOM objects
+
+### Representation of a `Mapping` object
+The RDF type of a `Mapping` object is `owl:Axiom`.
+
+If the `Mapping` object has a `record_id` slot, then the value of that
+slot is used as the named RDF resource that represents the object (and
+consequently, that slot MUST NOT be represented using the [general
+rules](#sssom-slots) for the representation of slots as defined above).
```
matentzn: maybe better to phrase this postively, like "rules don't apply".

src/docs/spec-formats-rdf.md
```diff
@@ -0,0 +1,366 @@
...
+consequently MUST NOT be represented using the [general
+rules](#sssom-slots) for the representation slots as defined above).
+
+The `curie_map` slot MUST NOT be represented using the [general
+rules](#sssom-slots). Instead, if it is needed it MUST be represented
+using whatever mechanism is provided by the concrete RDF serialisation
+format (e.g. `@prefix` declarations in [RDF
+Turtle](https://www.w3.org/TR/turtle/) or [RDF
+TriG](https://www.w3.org/TR/trig/), or `xmlns` namespace declarations in
+[RDF/XML](https://www.w3.org/TR/rdf-syntax-grammar/)).
```
matentzn, line 173: I am a bit against any relationship between curie map and the various  RDF prefix syntaxes you list. cc @cthoyt 

My main concern is this: two of the most important serialisations (RDF/XML, OWL/XML) can't even accurately represent a sssom:curie_map. Because in these XML serializations the prefix system is hooked into the XML namespacing system, the local identifier part has severe syntactic constraints. In particular, it must correspond to an NCNAME, which means it MUST start with a letter (not a number, so you cant actually represent UBERON:123 in RDF XML). 

My vote is to represent the prefix map using the SHACL prefixmap:

```
<?xml version="1.0"?>
<rdf:RDF
    xmlns:sh="http://www.w3.org/ns/shacl#"
    xmlns:rdf="http://www.w3.org/1999/02/22-rdf-syntax-ns#"
    xmlns:xsd="http://www.w3.org/2001/XMLSchema#"
    xmlns:ex="http://example.org/"
    xmlns:foaf="http://xmlns.com/foaf/0.1/"
    xml:base="http://example.org/shapes">

  <!-- Define the prefix map -->
  <sh:NodeShape rdf:about="#MyPrefixMap">
    <sh:prefixes>
      <sh:PrefixDeclaration>
        <sh:prefix>ex</sh:prefix>
        <sh:namespace>http://example.org/</sh:namespace>
      </sh:PrefixDeclaration>
      <sh:PrefixDeclaration>
        <sh:prefix>foaf</sh:prefix>
        <sh:namespace>http://xmlns.com/foaf/0.1/</sh:namespace>
      </sh:PrefixDeclaration>
    </sh:prefixes>
  </sh:NodeShape>

</rdf:RDF>
```

This ensures 100% faithful roundtripping.

src/docs/spec-formats-rdf.md
```diff
@@ -0,0 +1,366 @@
...
+
+> Non-normative notes
+>
+> 1. The CURIE map may not be needed at all if all named resources and
+>    predicates are always serialised as full-length IRIs.
```
matentzn, line 178: This is only true if we decide that SSSOM-RDF is export only format. I know we said this for SSSOM-OWL, forgot where we ended up with RDF.

src/docs/spec-formats-rdf.md
```diff
@@ -0,0 +1,366 @@
...
+> 2. If at least some named resources or predicates are serialised as
+>    CURIEs, the RDF requirement that all used prefix names must be
+>    declared (using the appropriate mechanism for the chosen concrete
+>    syntax) takes precedence over the possibility of omitting the
+>    declarations of prefix names that are considered
+>    [built-in](spec-intro.md#iri-prefixes) in the context of SSSOM.
```
matentzn: I would compact this sentence which has many redundant parts (the RDF requirement that all used prefix names must be declared) to:

> Prefixes considered [built-in](spec-intro.md#iri-prefixes) by other serialisations MUST be rendered using their fully qualified name (IRI / CURIE).

src/docs/spec-formats-rdf.md
matentzn: Also, is this true for `mappings` slot as well? Probably not right?

src/docs/spec-formats-rdf.md
```diff
@@ -0,0 +1,366 @@
...
+
+### Representation of a `ExtensionDefinition` object
+The RDF type of a `ExtensionDefinition` object is
+`sssom:ExtensionDefinition`.
+
+A `ExtensionDefinition` object has no identifier of any kind and is
+always represented by a blank node.
+
+## Special considerations for serialising to RDF
```
matentzn, line 192: mark as not normative?

src/docs/spec-formats-rdf.md
```diff
@@ -0,0 +1,366 @@
...
+format (e.g. SSSOM/TSV), or if it is only intended to be used by
+“generic”, non-SSSOM-aware RDF applications.
+
+Depending on that intended usage (if it is known), implementations may
+adopt slightly different behaviours as described in the following
+subsections.
+
+### Serialisations of identifiers
+If the serialisation is intended to be convertible back to another SSSOM
+format (especially the SSSOM/TSV format), implementations MUST
```
matentzn: Will comment on this after curie_map representation is sorted.

src/docs/spec-formats-rdf.md
gouttegd: > for each value {v1,v..,vn} represented by a set of individual triples {a,b,v1; a,b,v2,...a,b,vn}.

Does not sound any clearer to me. On the contrary, it sounds like each value is represented by a set of triples, which is certainly not the case.

> Also, is this true for mappings slot as well? Probably not right?

Of course it is. Mappings are represented as follows:

```ttl
?mappingset sssom:mappings [ a owl:Axiom ;
                               owl:annotatedSource ...
                           ] ,
                           [ a owl:Axiom ;
                               owl:annotatedSource ...
                           ] .
```

which fits the description for multi-valued slots: one triple per value.

This is what SSSOM-Py has always done, so I had assumed you were fine with that.

Let me guess: you are no longer happy with that and want to radically change the format?

src/docs/spec-formats-rdf.md
gouttegd: First, I would strongly advise against using `sssom:NonRelativeURI` as the `type_hint` for an extension slot.

That type is a hack that is only needed because (1) some people have strong opinion on what a `xsd:anyURI` value should look like (in particular, it should not store a relative URI reference) and (2) SSSOM tends to use URI-typed slots where an EntityReference would have been preferable given the intended usage of the concerned slots.

I would very much like this hack to remain purely internal to SSSOM. If people have URI-typed values, then

* either those values are semantic identifiers, and they should use `linkml:uriOrCurie`;
* or those values are merely, well, URIs, and there should be no reason for them not to use `xsd:anyURI`. Everybody else outside of SSSOM has been happy with that type for the past 20 years.

Using a `sssom:NonRelativeURI` type in the RDF representation of a mapping set is likely to do more harm than good, if only because, outside of SSSOM, nobody will have the slightest idea of what that type is – whereas any RDF person is likely to be familiar with `xsd:anyURI`.

Second: Simplicity and avoidance of special cases. It’s much easier, both for the spec writer and for the spec implementer, if all extension values can be treated the same way. There is _one_ special case (`uriOrCurie`) for semantic identifiers, but beyond that, the value of the `type_hint` does not matter.

src/docs/spec-formats-rdf.md
matentzn: No, it seems I misunderstood  

```
multi-valued slots are represented by as many
triples as there are values, each value being the object of one triple.
```

I thought this literally meant:

```
for each value {v1,v..,vn} represented by a set of individual triples {a,b,v1; a,b,v2,...a,b,vn}.
```

Which is why I was confused.

Maybe just add an example here?

src/docs/spec-formats-rdf.md
gouttegd: This is why there is a subsection ”serialisation of identifiers” in the “Special considerations” section.

The SSSOM/RDF format can be _both_ a read/write format that is equivalent to SSSOM/TSV or SSSOM/JSON (meaning that can roundtrip between all those formats without loss of information) _and_ an export format. It all depends on what you want to do with the output file – something the spec cannot know in advance.

In a sense, SSSOM/RDF is a “à la carte” format. You want to preserve the ability to roundtrip back to SSSOM/TSV? You can, just make sure to include the CURIE map and any extension definitions. You are not interested in being able to come back (say, because all you want is to load the set into a graph database and forget about it)? Then you don’t have to worry about the CURIE map or extension definitions at all.

src/docs/spec-formats-rdf.md
gouttegd: > My main concern is this: two of the most important serialisations (RDF/XML, OWL/XML) can't even accurately represent a sssom:curie_map. Because in these XML serializations the prefix system is hooked into the XML namespacing system, the local identifier part has severe syntactic constraints. In particular, it must correspond to an NCNAME, which means it MUST start with a letter (not a number, so you cant actually represent UBERON:123 in RDF XML).

Point taken, but that’s a limitation of RDF/XML. Other prefix-supporting formats such as Turtle don’t have that limitation.

You want to preserve the CURIE map accurately and preserve the ability to roundtrip? Then don’t use RDF/XML.

> My vote is to represent the prefix map using the SHACL prefixmap:

You are joking. You are joking, right?

You are not seriously proposing something _completely different_ from what we currently have? Something for which we don’t have the inkling of the beginning of support in neither SSSOM-Py nor SSSOM-Java?

Because if you are not joking, I give up.

Design your dream format all you want, and give me a sign when you’ll be done moving the goalposts from one day to the next.

matentzn: > You are joking. You are joking, right?

I did not expect such a strong disagreement :P 

Alright, instead of this suggestion, we should then make SSSOM-RDF explicitly write only format.. I am totally fine with this as well, and we don't need to concern ourselves then _at all_ with the curie_map.. This way we can circumvent the limitations outlined above?

gouttegd: > I did not expect such a strong disagreement :P

Try sparing some thoughts for the people who have to implement your last-minute ideas that you sprout out of nowhere, maybe you’ll understand.

Two days ago you were reluctant to just changing the predicates to use to represent the core triple (e.g. changing from `owl:annotatedSource` to `sssom:subject_id`, which would have been a much smaller change and easier change than what you’re suggesting here). I thought we were in agreement that the SSSOM/RDF format should be as close as possible to what we already have, both to avoid needlessly breaking things and to keep the work needed to make SSSOM-Py support the updated format minimal (given that we have very little developer time available to work on SSSOM-Py). The spec I proposed is tailored for that; it deviates only minimally from the existing, _de facto_ SSSOM/RDF format that SSSOM-Py has been producing since forever; it is already completely supported by SSSOM-Java and I am hopeful that SSSOM-Py could be made to support it relatively easily (precisely due to the minimal deviations). You were fine with the [initial proposal](https://github.com/mapping-commons/sssom/discussions/454#discussioncomment-13655841), on which you said that you had merely “minor stylistic comments”.

And now, you’re here suggesting that we should in fact do something completely different, that has never even been casually mentioned in any discussion related to the SSSOM/RDF format. Because all of a sudden you are concerned about roundtripping back from RDF/XML, even though the RDF/XML produced by SSSOM-Py has never been roundtrippable and AFAIK nobody has ever complained about that.

So yeah, I disagree with your proposal, because it contradicts everything that you _seemed_ to want just two days ago.

So what do you want _now_?

A. A SSSOM/RDF format that is minimally different from what we already have, that can be supported rapidly (that is _already_ supported by one implementation), but that (oh, the horror!) does not guarantee that a set written in RDF/XML can be roundtripped back to another SSSOM format?

B. A SSSOM/RDF format that is a clean break from the existing stuff, that will initially not be supported by any implementation (and in fact I doubt it will _ever_ be implemented by SSSOM-Py, given the lack of activity on that implementation)?

If you want B, fine. But then I’ll leave _you_ to design the format. I won’t get involved in any of it, I’ll just wait until you have designed the perfect format of your dream, and then I promise I’ll do my best to implement it.

gouttegd: > we should then make SSSOM-RDF explicitly write only format.. 

Now you are throwing the baby with the bathwater. Just because the RDF/XML concrete serialisation may not guarantee that the prefix map is preserved does not mean that we should give up on SSSOM/RDF being a read/write format.

Again: As currently written, the spec does allow a SSSOM/RDF set to be fully converted back to any other SSSOM format, provided that:

* you do not serialise into RDF/XML;
* you serialise identifiers as CURIEs and make sure to include the appropriate prefix declarations.

As I said in another comment below, I wrote the spec to be flexible (“à la carte”): if you want the ability to roundtrip between RDF and another format, you can have it; if you are not interested in that ability, you can ignore it.

src/docs/spec-formats-rdf.md
gouttegd: Start by stating exactly what you expect from the SSSOM/RDF spec. I thought I knew, but in fact I cannot know it since it seemingly changes every other day.

src/docs/spec-formats-rdf.md
gouttegd: > * you do not serialise into RDF/XML;
> * you serialise identifiers as CURIEs and make sure to include the appropriate prefix declarations.

In fact even if you do serialise into RDF/XML and write identifiers as IRIs, you will _still_ be able convert back to SSSOM/TSV, _unless_ you happen to run your RDF/XML file into a RDF processor that decides to strip away any unused namespace declarations. Not sure if that is a common behaviour among RDF tools, but Jena and RDFLib do _not_ seem to do it – they are happy to let unused namespace declarations pass through unchanged.

src/docs/spec-formats-rdf.md
```diff
@@ -0,0 +1,366 @@
...
+      owl:annotatedProperty skos:exactMatch;
+      owl:annotatedSource KF_FOOD:F002;
+      owl:annotatedTarget FOODON:00003348;
+      EXT:isFooable false;
+      sssom:confidence 1.0E0;
+      sssom:mapping_justification semapv:ManualMappingCuration;
+      sssom:object_label "Gala apple (whole)";
+      sssom:subject_label "gala"
+    ] .
+```
```
cthoyt, line 451: I would expect that the triples represented by the axioms also to show up somewhere in the RDF

````suggestion
KF_FOOD:F001 skos:exactMatch FOODON:00002473 .
KF_FOOD:F002 skos:exactMatch FOODON:00003348 .
```
````

though it's not clear what to do for negated triples

File: src/docs/spec-formats-rdf.md [hunk 12]
```diff
@@ -0,0 +1,366 @@
...
+For example, implementations MAY accept
+
+```ttl
+?mapping sssom:predicate_modifier "Not"^^xsd:string .
+```
+
+as an alternative to
+
+```ttl
+?mapping sssom:predicate_modifier sssom:NegatedPredicate .
```
cthoyt, line 351: why not just decide on one and say "this is the standard"?

gouttegd: That’s what we did. This:

```ttl
?mapping sssom:predicate_modifier sssom:NegatedPredicate .
```

**is** the standard.

But the decision to standardize that form has only been made a few weeks ago. Before that, both SSSOM-Java and SSSOM-Py have been producing the string literal form (SSSOM-Java for the past 8 months – since version 1.1, which introduced RDF support – and SSSOM-Py for as long as it has existed). In fact SSSOM-Py _still_ produces the string literal form to this day.

So for backwards compatibility (which is the entire point of this section, “compatibility with pre-standard representations”), implementations MAY support the old string literal form, even though it is _not_ the standard form.

src/docs/spec-formats-rdf.md
gouttegd: They appear, but only as reified OWL axioms.

This has been the RDF output produced by SSSOM-Py since the beginning.

src/docs/spec-formats-rdf.md
matentzn: Lets tackle the NCname issue with a comment in the documentation: XML formats might not be able to roundtrip

(@gouttegd says: longest URI extension will win)

Maybe just document a rule on conflicting uri prefixes for roundtrip

- cant control curiefication in RDF based formats
- add "longest uri expansion" assumption to canonicalisation and refer to canonicalisation as "preprocessing" for RDF generation.

src/docs/spec-formats-rdf.md
matentzn: Withdrawing this comment as we have decided to stick with the current way of doing things in sssom implementations for now.

src/docs/spec-formats-rdf.md
matentzn: Ok!

src/docs/spec-formats-rdf.md
matentzn: Makes sense.

src/docs/spec-formats-rdf.md
matentzn: Its a valid question though. Without the direct triple, triples stores might not be able to return all terms mapped to `?x` with a simple triple pattern - this is only possible because in many of the pipelines I had build that are doing gathering dust, I loaded the sssom file into robot and saved it, which I believe automatically injects that triple. 

In the OWL serialisation it seems I have [injected them](https://github.com/mapping-commons/sssom-py/blob/ab69eeb9f0611d8e5d0c436f4ecf556ded8bea97/src/sssom/writers.py#L215) in sssom py:
https://github.com/mapping-commons/sssom-py/blob/master/tests/validate_data/cob-to-external.tsv.owl

gouttegd: > Without the direct triple, triples stores might not be able to return all terms mapped to ?x with a simple triple pattern

If that is needed, once the set has been exported to RDF it shouldn’t be hard to process it with some SPARQL to construct a `?subject_id ?predicate_id ?object_id` triple from every `?mapping owl:annotatedSource ?subject_id ; owl:annotatedPredicate ?predicate_id ; owl:annotatedTarget ?object_id` set of triples, before loading it into a triple store.

I wouldn’t mind adding that as a OPTIONAL behaviour for RDF writer, on the condition that it is really optional – that is, if those `?subject_id ?predicate_it ?object_id` triples are absent the set must still be accepted by a SSSOM/RDF reader, as long as the `?mapping owl:annotatedSource ?subject_id ; owl:annotatedPredicate ?predicate_id ; owl:annotatedTarget ?object_id` triples are present.

Something like:

> A SSSOM/RDF writer MAY additionally inject for every mapping record a triple of the form `?subject_id ?predicate_id ?object_id`. A SSSOM/RDF reader MUST NOT expect the presence of such triples.

If we do allow that, this raises the question (as hinted by @cthoyt) of what to do about negated mappings. Two possibilities:

(A) Don’t care. Negated mappings are treated in the same way as any other mappings. If users don’t want `?subject_id ?predicate_id ?object_id` triples for negated mappings, it’s up to them to filter out negated mappings before exporting the set to RDF.

(B) Explicitly exclude negated mappings, as in

> A SSSOM/RDF writer MAY additionally inject for every mapping record a triple of the form `?subject_id ?predicate_id ?object_id`, only for mapping records that do not have a `predicate_modifier` of `Not`.

gouttegd: There is also the question of mappings with `sssom:NoTermFound`, e.g.

| record_id | subject_id | predicate_id | object_id | object_source |
| ----------| ---------- | ------------ | --------- | ------------- |
| MYMAP:1   | HP:1234    | skos:exactMatch | sssom:NoTermFound | obo:doid.owl |

Rendering them as

```ttl
MYMAP:1 a owl:Axiom ;
           owl:annotatedSource HP:1234 ;
           owl:annotatedPredicate skos:exactMatch
           owl:annotatedTarget sssom:NoTermFound ;
           sssom:object_source obo:doid.owl .
```

should be perfectly fine, but do we also want a

```ttl
HP:1234 skos:exactMatch sssom:NoTermFound .
```

triple as well?

I’d say, we should do for them the same thing as we do for negated mappings (the two possibilities outlined in my previous message).

matentzn: I am now on board with:

> I wouldn’t mind adding that as a OPTIONAL behaviour for RDF writer, on the condition that it is really optional – that is, if those ?subject_id ?predicate_it ?object_id triples are absent the set must still be accepted by a SSSOM/RDF reader, as long as the ?mapping owl:annotatedSource ?subject_id ; owl:annotatedPredicate ?predicate_id ; owl:annotatedTarget ?object_id triples are present.

With regards to the special cases:

> I’d say, we should do for them the same thing as we do for negated mappings (the two possibilities outlined in my previous message).

I agree. Just because both are very different use cases I would favour it if implementations would be injecting triples to sssom:NoTermFound versus negated mappings based on _separate_ conditionals; I  would probably never add either one, but there may be some use cases to do so.

gouttegd: > I would favour it if implementations would be injecting triples to sssom:NoTermFound versus negated mappings based on separate conditionals

Please define what you mean by “favour”. For now, we are talking about a completely optional behaviour, as in:

> A SSSOM/RDF writer MAY additionally inject for every mapping record a triple of the form ?subject_id ?predicate_id ?object_id.

(Note the _MAY_.) Since writing the SPO triple is an optional behaviour, we cannot expect any “favour” from the implementations, which are completely free to (1) implement the behaviour or not and (2) to implement it in whatever way they want.

If you _want_ implementations to offer the option of injecting SPO triples, then you have to _require_ it in the spec, as in:

> SSSOM/RDF writers MUST allow to optionally inject for every mapping record a triple of the form ?subject_id ?predicate_it ?object_id.

Note the difference compared to the previous phrasing. Here it’s no longer an _optional feature_, it’s a _mandatory_ feature (whose _activation_ is optional, but the feature itself has to be present, it’s no longer left for implementers to decide whether they want to support it or not).

gouttegd: If we add to that the issue about `sssom:NoTermFound` and negated mappings, in the end you seem to want to require _three_ options:

(1) One option to enable/disable the production of SPO triple entirely:

> SSSOM/RDF writers MUST allow to optionally inject for every mapping record a triple of the form ?subject_id ?predicate_id ?object_id.

(2) One option to enable the production of SPO triples to `sssom:NoTermFound`:

> When the injection of ?subject_id ?predicate_id ?object_id triples is enabled, SSSOM/RDF writers MUST NOT inject such triples for records where the subject ID or the object ID (or both) is `sssom:NoTermFound`, unless this has been explicitly requested.

(3) One option to enable the production of SPO triples for negated mappings:

> When the injection of ?subject_id ?predicate_id ?object_id triples is enabled, SSSOM/RDF writers MUST NOT inject such triples for records where `predicate_modifier` is set to `Not`, unless this has been explicitly requested.

Frankly this seems way overkill to me. You even say yourself that you would probably never have a need for such options, so why _requiring_ from implementations that they offer them?

src/docs/spec-formats-rdf.md
gouttegd: > Prefixes considered [built-in](https://github.com/mapping-commons/sssom/pull/spec-intro.md#iri-prefixes) by other serialisations MUST be rendered using their fully qualified name (IRI / CURIE).

I am sorry but that sentence makes no sense. What does it mean to “render a prefix using its fully qualified name (IRI / CURIE)”?

What we need to express here is that “built-in” prefixes MUST be declared as any other prefixes (in effect, there is no such thing as a “built-in prefix” in the RDF world).

src/docs/spec-formats-rdf.md
gouttegd: It _is_ normative in my opinion. Implementations SHOULD consider those issues when deciding how to implement RDF serialisations.

There are non-normative explanations throughout the section.

src/docs/spec-formats-rdf.md
matentzn: > Frankly this seems way overkill to me. You even say yourself that you would probably never have a need for such options, so why requiring from implementations that they offer them?

Ok, I agree with the plan. As long as by default, (2) and (3) are NOT injected when the option to inject triples is implemented. So:

> (1) One option to enable/disable the production of SPO triple entirely:

SSSOM/RDF writers MAY allow to optionally inject for every mapping record a triple of the form ?subject_id ?predicate_id ?object_id.

I would keep your phrasing of (2) and (3) though because we really dont want these two cases to be materialised, even if an implementation chooses to implement direct injection?

gouttegd: > Ok, I agree with the plan. […] I would keep your phrasing of (2) and (3)

Err, with _what_ do you agree? In the end, you still want the _three_ options.

matentzn: I thought you meant to make implementing these options an optional possibility? Maybe state again what exactly you are in favour of documenting here. I thought you were just against "requiring" these options? I am just saying that:

1. If you add that option, dont just inject noTermFound triples by default

gouttegd: > I am just saying that: If you add that option, dont just inject noTermFound triples by default

OK, I must have misunderstood then, my bad.

The current proposition (that is, the current state of the PR) is exactly the following:

> ## Representation of mappings as “direct triples”
> 
> For every single mapping record in a set, implementations MAY additionally inject a single triple of the form:
> 
> ```
> ?subject_id ?predicate_id ?object_id .
> ```
> 
> If so, implementations SHOULD NOT inject such triples in the following cases:
>
> * when the record represents a literal mapping (that is, subject_type or object_type – or both – is set to rdfs literal);
> * when the record represents a negated mapping (that is, predicate_modifier is set to Not);
> * when the record represents an absence of match (that is, subject_id or object_id – or both – is set to sssom:NoTermFound).
> 
> In any case, a SSSOM/RDF reader MUST NOT expect the presence of such triples, and if they are present MUST NOT use them to construct mapping records.

Which does not correspond exactly to what you want, because “implementations MAY additionally inject” does not mandate that the behaviour has to be _optional_ – with that phrasing an implementation could very well decide that to _always_ inject the direct triples.

So I’ll rephrase that as:

> For every single mapping record in a set, implementations MAY additionally inject a single triple of the form:
> 
> ```
> ?subject_id ?predicate_id ?object_id .
> ```
> 
> If so, that behaviour MUST be optional.
>
> When that behaviour is enabled, implementations SHOULD NOT inject such triples in the following cases:
>
> * when the record represents a literal mapping (that is, subject_type or object_type – or both – is set to rdfs literal);
> * when the record represents a negated mapping (that is, predicate_modifier is set to Not);
> * when the record represents an absence of match (that is, subject_id or object_id – or both – is set to sssom:NoTermFound).
> 
> In any case, a SSSOM/RDF reader MUST NOT expect the presence of such triples, and if they are present MUST NOT use them to construct mapping records.

I think this matches what you want?

matentzn: Yes, I think this makes sense! Thank you! 🙏

gouttegd: OK, I will push the change later today.

src/docs/spec-formats-rdf.md
```diff
@@ -0,0 +1,422 @@
...
+
+The value MUST be represented as a `xsd:double` literal.
+
+#### For slots typed as an enumeration
+(e.g. `sssom_version`, `mapping_cardinality`, `subject_type`…)
+
+If the permissible values for the enumeration are defined in the LinkML
+model as having an associated `meaning` property, then the value MUST be
+represented as a named RDF resource with the indicated property.
+Otherwise, the value MUST be represented as a string literal.
```
matentzn:
```suggestion
Otherwise, the value MUST be represented as an `xsd:string` literal.
```

To maintain consistency with the descriptions above

src/docs/spec-formats-rdf.md
```diff
@@ -0,0 +1,422 @@
...
+represented.
+
+> Example:
+>
+> A `creator_id` slot with the values `https://example.org/people/0001`
+> and `https://example.org/people/0002` is represented by the following
+> two triples:
+>
+> ```ttl
+> ?object pav:authoredBy <https://example.org/people/0001> .
```
matentzn:
```suggestion
> ?object dcterms:creator <https://example.org/people/0001> .
```

src/docs/spec-formats-rdf.md
```diff
@@ -0,0 +1,422 @@
...
+> ?object pav:authoredBy <https://example.org/people/0002> .
```
matentzn:
```suggestion
> ?object dcterms:creator <https://example.org/people/0002> .
```

src/docs/spec-formats-rdf.md
matentzn: https://mapping-commons.github.io/sssom/creator_id/

src/docs/spec-formats-rdf.md
matentzn: I dont know what my July-self exactly was thinking here.

src/docs/spec-formats-rdf.md
matentzn: I will resolve this comment now, looks great

matentzn (commented):
Some cosmetic things, I will ask two reviewers to chime in so we can merge this asap!

This is huge work, thanks @gouttegd!! much appreciated!

cthoyt:
i'm going to run prettier on the markdown before reviewing it, after damien has a chance to address your suggestions

src/docs/spec-formats-rdf.md
gouttegd: Doesn’t seem strictly necessary to me as it should be obvious (when we are talking about RDF) that a “string literal” is a `xsd:string` literal, but I can’t object to the consistency argument. :)

The “**an**” before the `xsd:string` looks very weird to me though, and for consistency with other instances of “a `xsd:...`” I’m going to reject it.

src/docs/spec-formats-rdf.md
gouttegd: Yep, my bad, I confused with `author_id`.

src/docs/spec-formats-rdf.md
cthoyt: if you say it out loud, you would say `an X-S-D string literal` because X starts with a vowel sound, but agreed, it's a bit awk in writing 😁

src/docs/spec-formats-rdf.md
```diff
@@ -0,0 +1,429 @@
...
+## Representation of slots
+
+A metadata slot on any given SSSOM object (such as a `Mapping` or a
+`MappingSet`) MUST be represented as a RDF triple where:
+
+- the subject is the resource representing the SSSOM object;
+- the predicate is either:
+  - the property indicated by the `URI` field in the LinkML description of the
+    slot, if such a field is present;
+  - or a property constructed by concatenating the `https://w3id.org/sssom/`
```
cthoyt, line 32: is there a situation where this actually happens? Shouldn't the LinkML schema be explicit and exhaustive?

src/docs/spec-formats-rdf.md
```diff
@@ -0,0 +1,429 @@
...
+(e.g. `mapping_set_confidence`, `confidence`, `similarity_score`)
+
+The value MUST be represented as a `xsd:double` literal.
+
+#### For slots typed as an enumeration
+
+(e.g. `sssom_version`, `mapping_cardinality`, `subject_type`…)
+
+If the permissible values for the enumeration are defined in the LinkML model as
+having an associated `meaning` property, then the value MUST be represented as a
```
cthoyt, line 70: can you please give an explicit example?

src/docs/spec-formats-rdf.md
```diff
@@ -0,0 +1,429 @@
...
+named RDF resource with the indicated property. Otherwise, the value MUST be
+represented as a `xsd:string` literal.
+
+#### For slots typed as a SSSOM object
+
+(e.g. `mappings`, `extension_definitions`)
+
+The value MUST be represented as a RDF resource. Whether the resource is named
+(IRI) or not (blank node) will depend on the type of the object, see the
+[section on representing SSSOM objects](#sssom-objects) below for details.
```
cthoyt, line 98: this is a bit confusing why it's documented here but not in full, can you make an explicit enumeration here of the three things this applies to an an link to each?

```suggestion
The value MUST be represented as a RDF resource. Whether the resource is named
(IRI) or not (blank node) will depend on the type of the object, see the following sections:

1. [mappings](#representation-of-a-mapping-object)
2. [mapping set)(#representation-of-a-mappingset-object)
3. [extension definition](#representation-of-a-extensiondefinition-object)
```

src/docs/spec-formats-rdf.md
```diff
@@ -0,0 +1,429 @@
...
+declared prefixes.
+
+> Non-normative explanation
+>
+> This is because, if all identifiers are serialised as full-length IRIs, then
+> even if the RDF file includes prefix declarations, they may be stripped away
+> by a RDF reader, since they are not needed. And without those prefix
+> declarations, it would not be possible to serialise the set back as a
+> SSSOM/TSV file (remember that the SSSOM/TSV format _requires_ that identifiers
+> be serialised as CURIEs).
```
cthoyt, line 217: shade thrown at CURIEs :p. Let's extend this with a sentence with some context like 

```suggestion
> SSSOM/TSV file (remember that the SSSOM/TSV format _requires_ that identifiers
> be serialised as CURIEs to support the easy understanding on its content).
```

src/docs/spec-formats-rdf.md
```diff
@@ -0,0 +1,429 @@
...
+```ttl
+?subject_id ?predicate_id ?object_id .
+```
+
+If so, that behaviour MUST be optional.
+
+When that behaviour is enabled, implementations SHOULD NOT inject such triples
+in the following cases:
+
+- when the record represents a literal mapping (that is, `subject_type` or
```
cthoyt, line 278: oh man, every time I remember this is actually a part of SSSOM, I cry

src/docs/spec-formats-rdf.md
```diff
@@ -0,0 +1,429 @@
...
+>
+> It is recommended not to inject such direct triples for literal mapping
+> records, even if they do have a `subject_id` and a `object_id`, because by
+> definition the subject and/or the object of such records is not an
+> identifiable semantic entity and has no business being represented in a RDF
+> graph.
+>
+> It is recommended not to inject such direct triples for negated mapping
+> records because they would seem to convey a meaning that is the exact opposite
+> of what the records mean.
```
cthoyt, line 304: Thanks, this is super important. I think it begs the more general question of if there's a standard for representing "negated" triples in RDF. Maybe we can link to external information with more explanation about why this doesn't exist, if it doesn't (and if it does, maybe we should use it)

I don't think it exists after some looking around

src/docs/spec-formats-rdf.md
```diff
@@ -0,0 +1,429 @@
...
+>
+> It is recommended not to inject such direct triples for no-match mapping
```
cthoyt, line 306: Musing: we might consider for the future how to represent the fact that there's no match in a different way

src/docs/spec-formats-rdf.md
```diff
@@ -0,0 +1,429 @@
...
+
+In the name of backward compatibility, implementations MAY support the
+alternative rules described in the following subsections when deserialising from
+RDF.
+
+Implementations MUST NOT follow these rules when serialising to RDF.
+
+### Representation of slots typed as `sssom:NonRelativeURI`
+
+Implementations MAY accept a value represented as a `xsd:anyURI` literal.
```
cthoyt, line 323: example please of what worked before and doesn't work anymore

src/docs/spec-formats-rdf.md
gouttegd: Err, have you looked at the SSSOM schema? It happens _everywhere_. Most of the slots do _not_ have an explicit `URI` field – because for most of the slots there isn’t a readily available property in a pre-existing vocabulary.

src/docs/spec-formats-rdf.md
gouttegd: Mildly against it ① as it means there will be one more place to update should we ever add a new type of object, and ② I don’t think it is that confusing. It is expected, and maybe even unavoidable, that a specification document should contain “forward references” (references to things that are detailed later), implementers should not be confused by that.

src/docs/spec-formats-rdf.md
gouttegd: Sorry, what? This is a factual statement, how is that a “shade thrown at CURIEs”?

And I believe this factual statement is necessary because people regularly forget the SSSOM/TSV-specific requirement that identifiers MUST be in CURIE form – even the very people who should know very well about that requirement, as in [this discussion](https://github.com/mapping-commons/sssom/issues/399#issuecomment-2493913177).

Besides, your added sentence is incorrect. The fact that SSSOM/TSV absolutely requires identifiers to be in CURIE form has nothing to do with “supporting the easy understanding on its contents”.

First because it is dubious that `FBbt:00004508` is easier to understand that `http://purl.obolibrary.org/obo/FBbt_00004508` – personally I’d say it’s _worse_, because now you need to look up what `FBbt` stands for, whereas the full-length IRI is self-sufficient.

Second, even if CURIEs were actually easier to understand, SSSOM/TSV could very well have simply _encouraged_ the use of CURIE forms, instead of _requiring_ it (as in, “identifiers in SSSOM/TSV files SHOULD be CURIE form”, instead of “identifiers in SSSOM/TSV files MUST be in CURIE form”).

The real reason SSSOM/TSV absolutely requires the use of CURIEs is, if I recall correctly, because SSSOM-Py developers didn’t want to have to deal with the possibility that a SSSOM/TSV could contain both full-length identifiers and CURIEs. Notably because this would have required, either the use of some heuristics to infer the form of an identifier, or the use of some syntactic sugar to distinguish between an IRI and a CURIE, like what is done in most RDF syntaxes – e.g. IRI in angled brackets vs “naked” CURIEs –, which nobody wanted for some reason.

But regardless all that, there is no need for the RDF specification to elaborate on _why_ the SSSOM/TSV requires the use of CURIEs. People simply need to be reminded that it is the case, because they frequently forget about it and this has some consequences if you want to be able to re-convert a SSSOM/RDF mapping set back to the SSSOM/TSV format.

src/docs/spec-formats-rdf.md
gouttegd: Well, the previous solution (the so-called “literal profile”) was even worse.

src/docs/spec-formats-rdf.md
gouttegd: The only thing I am aware of (which, does it even need to be said, does not mean a lot! :D ) is the concept of [RDF Surfaces](https://www.semantic-web-journal.net/system/files/swj3799.pdf).

Basically, the idea (as far as I understand it) is that some RDF triples could live within a “negated surface”, a part of the RDF graph that only contains negated assertions. This would look like this:

```n3
@prefix log: <http://www.w3.org/2000/10/swap/log#> .
@prefix FBbt: <http://purl.obolibrary.org/obo/FBbt_> .
@prefix UBERON: <http://purl.obolibrary.org/obo/FBbt_> .
@prefix skos: http://www.w3.org/2004/02/skos/core#> .

(_:x) log:onNegativeSurface {
  FBbt:00004508 skos:exactMatch UBERON:0000056 .
} .
```

Of note, however:

① The paper is still [under review](https://www.semantic-web-journal.net/content/rdf-surfaces-enabling-classical-negation-and-first-order-expressivity-semantic-web), so it certainly doesn’t look like it’s ready for prime-time.

② Not sure how much of that is actually usable in RDF. The examples in the paper (and the example above, which is derived from them) use N3 because it allows to express [formulae](https://www.w3.org/TeamSubmission/n3/#Quoting) (the list of statements enclosed within `{` `}`), something that does not exist in RDF.

src/docs/spec-formats-rdf.md
gouttegd: The time to discuss that was before the 1.0 release last year. Now, even if we come up with another way to represent the absence of match, we will have to keep the `sssom:NoMatchFound` mechanism for compatibility anyway.

src/docs/spec-formats-rdf.md
gouttegd: Fine. But frankly, if someone needs an example to understand the difference between a named RDF resource and a `xsd:anyURI` literal, then maybe that person should not come anywhere near the code of a SSSOM/RDF parser.

src/docs/spec-formats-rdf.md
cthoyt: I meant making our own RDF idiom for it

gouttegd: Is there really a need for it?

Given the following mapping:

```
subject_id      predicate_id      object_id
FBbt:00004568   skos:exactMatch   sssom:NoTermFound
```

The fact `FBbt:00004568` has no match _is_ already expressed in the RDF serialisation:

```ttl
[] a owl:Axiom .
   owl:annotatedSource FBbt:00004568 ;
   owl:annotatedProperty skos:exactMatch ;
   owl:annotatedTarget sssom:NoTermFound .
```

Yes, it does not appear as a “direct triple”, but I fail to see how is that a problem.

The point of “direct triples” is a convenience for RDF consumers, so that they can quickly find the mappings they want by simple queries over the RDF graph. Like, they want all the exact matches to FBbt:00004568? They just need to get all the triples where `FBbt:00004568` is the subject and `skos:exactMatch` is the predicate – something that, with most RDF libraries, is the matter of a single function call. And if that query does not return anything (which would be the case in our example here), well, they know that there is no match.

src/docs/spec-formats-rdf.md
cthoyt: fine for me

src/docs/spec-formats-rdf.md
cthoyt: thanks for the note, this might be useful later.

src/docs/spec-formats-rdf.md
cthoyt: okay, that makes sense

src/docs/spec-formats-rdf.md
cthoyt: Alright, we can skip this suggestion

cthoyt (dismissed):

src/docs/spec-formats-rdf.md
```diff
@@ -0,0 +1,459 @@
...
+format that may be used to serialise the model.
+
+It is RECOMMENDED that implementations support reading and writing a SSSOM set
+from and to the [RDF Turtle](https://www.w3.org/TR/turtle/) format at least.
+They MAY support any other RDF concrete format (e.g. RDF/XML, TriG, N-Triples,
+etc.).
+
+This specification does not mandate how a concrete RDF syntax is to be used. For
+example, if the RDF syntax allows named resources and predicates to be
+serialised as either IRIs or CURIEs, if is left at the discretion of the
```
ehartley:
```suggestion
serialised as either IRIs or CURIEs, it is left to the discretion of the
```

src/docs/spec-formats-rdf.md
```diff
@@ -0,0 +1,459 @@
...
+>    the RDF requirement that all used prefix names must be declared (using the
+>    appropriate mechanism for the chosen concrete syntax) takes precedence over
+>    the possibility of omitting the declarations of prefix names that are
+>    considered [built-in](spec-intro.md#iri-prefixes) in the context of SSSOM.
+
+### Representation of a `ExtensionDefinition` object
+
+The RDF type of a `ExtensionDefinition` object is `sssom:ExtensionDefinition`.
+
+A `ExtensionDefinition` object has no identifier of any kind and is always
```
ehartley:
```suggestion
### Representation of an `ExtensionDefinition` object

The RDF type of an `ExtensionDefinition` object is `sssom:ExtensionDefinition`.

An `ExtensionDefinition` object has no identifier of any kind and is always
```

src/docs/spec-formats-rdf.md
```diff
@@ -0,0 +1,459 @@
...
+
+Conversely, they SHOULD be included if the set is intended to be convertible
+back to another SSSOM format.
+
+> Non-normative explanation
+>
+> The whole point of an extension definition in SSSOM is to provide (1) a
+> property that confers some meaning to the extension, and (2) the type of the
+> expected values. In RDF, as described [above](#extension-slots), those two
+> bits of information are already contained in the triple that represent the
```
ehartley:
```suggestion
> bits of information are already contained in the triple that represents the
```

src/docs/spec-formats-rdf.md
```diff
@@ -0,0 +1,459 @@
...
+> Non-normative explanations
+>
+> Such “direct triples” are merely a convenience for downstream RDF
+> applications, allowing them to find a direct link (as a single triple) between
+> the subject and the object of a mapping, without having to construct such a
+> link by following the `owl:annotatedSource`, `owl:annotatedProperty`, and
+> `owl:annotatedTarget` triples.
+>
+> It is recommended not to inject such direct triples for literal mapping
+> records, even if they do have a `subject_id` and a `object_id`, because by
```
ehartley:
```suggestion
> records, even if they do have a `subject_id` and an `object_id`, because by
```

ehartley (approved):
There were a few minor typos, but otherwise I found the specification documentation to be really clear.

PR #470, opened 2025-07-16
Remove required/optional flag from slot pages

### LINKED ISSUE #468: Documentation templates make too many assumptions
Consider the [generated documentation](https://mapping-commons.github.io/sssom/type_hint/) of the `type_hint` slot. It includes this automatically generated bit:

> Required?: Optional (You can add this element to the mapping or mapping set, but it is not required.)

which obviously assumes that the slot is one of the slots of the `Mapping` or `MappingSet` classes. But it is not! `type_hint` is a slot of the [ExtensionDefinition class](https://mapping-commons.github.io/sssom/ExtensionDefinition/) (as indicated by the “Applicable to:” line).

More generally, the documentation template for slots (`src/doc-templates/slot.md.jinja2`) systematically assumes that any slot always belongs to either the `Mapping` or the `MappingSet` class, completely ignoring any other class in the model.

For example, the [page for the `homepage` slot](https://mapping-commons.github.io/sssom/homepage/) likewise includes the same bit indicating that the slot may be added to a mapping or a mapping set, even though that slot is specific to the `MappingRegistry` class.

matentzn:
As pointed out by @gouttegd in https://github.com/mapping-commons/sssom/issues/468, this template makes too many assumptions. A slot is not _generally_ optional or required, it is optional or required in the context of a specific class.

Fixes #468

### DIFF
--- a/src/doc-templates/slot.md.jinja2
+++ b/src/doc-templates/slot.md.jinja2
@@ -40,9 +40,6 @@
 
 * **Range**: {{gen.link(element.range)}}. The range of the element is the type of the value that can be assigned to it.
 {% if element.multivalued %}* **Multivalued**: {{ element.multivalued }}. If the element is multivalued, more than one value can be attached to the same field. In the TSV format, these are `|` separated.{% endif %}
-* **Required?**: {% if element.required %}Required (element has to be added to the mapping or mapping set.)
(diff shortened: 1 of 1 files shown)


gouttegd (approved):
Thanks!

ehartley (approved):

PR #472, opened 2025-07-18
Make curation_rule_* slots propagatable

### LINKED ISSUE #464: Questions about curation_rule
[curation_rule](https://mapping-commons.github.io/sssom/curation_rule/) and [curation_rule_text](https://mapping-commons.github.io/sssom/curation_rule_text/) are applicable to individual Mappings

**First question:** if both are given, do they refer to the same entity (unlike `author_id` and `author_label`)? The description (2) seems to implies this:

>  The textual representation of curation rule is intended to be used in cases where (1) the creation of a resource is not practical from the perspective of the mapping_provider and (2) as an additional piece of metadata to augment the curation_rule element with a human readable text. 

So having curation_rule `x:a|x:b` and curation_rule_text `A`, there would be two rules: `x:a` with text `A` and another rule `x:b` without text.

**Second question:** We collect sets of mappings that follow the same rule, so the rule better fits to the MappingSet. In particular guidelines which elements to map with priority and which elements not to map (e.g. if parent elements have already been mapped) are nothing that fits to the level of indivdual mappings. So I'd like to extend the curation_rule and curation_rule_text elements to also be applicable to MappingSet or make it a propagatable slot.

matentzn:
- Make curation_rule_* slots propagatable 
- Remove a confusing part of the curation rule description.

Resolves #464 

- [ ] ~tests have been added/updated (if applicable)~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 - Add `mapping_tool_id` slot to the `Mapping` and `MappingSet` classes ([issue](https://github.com/mapping-commons/sssom/issues/449)).
 - Add `record_id` slot to the `Mapping` class ([issue](https://github.com/mapping-commons/sssom/issues/359)).
 - Change all URI-typed slots to clarify that they expect _non-relative_ URIs as values ([issue](https://github.com/mapping-commons/sssom/issues/448)).
+- `curation_rule` and `curation_rule_text` are made propagatable ([issue](https://github.com/mapping-commons/sssom/issues/464)).
 
 ## SSSOM version 1.0.0
 
--- a/examples/schema/curation_rule-propagated.sssom.tsv
+++ b/examples/schema/curation_rule-propagated.sssom.tsv
@@ -1,0 +1,16 @@
+#curie_map:
+#  HP: http://purl.obolibrary.org/obo/HP_
+#  MP: http://purl.obolibrary.org/obo/MP_
+#  orcid: https://orcid.org/
+#  DISEASE_MAPPING_COMMONS_RULES: https://w3id.org/sssom/commons/disease/curation-rules/
+#mapping_set_id: https://w3id.org/sssom/commons/examples/curation_rule.sssom.tsv
+#mapping_set_description: "This example illustrates how to express that all mapppings in a mapping set have been curated according to a specific curation rule"
+#license: "https://creativecommons.org/publicdomain/zero/1.0/"
+#creator_id: orcid:0000-0002-7356-1779
+#mapping_provider: "https://w3id.org/sssom/core_team"
+#curation_rule: DISEASE_MAPPING_COMMONS_RULES:MPR2
+#comment: This is an example file for the SSSOM for illustration only. Its contents are entirely fabricated.
+subject_id	predicate_id	object_id	mapping_justification
+HP:0009124	skos:exactMatch	MP:0000003	semapv:ManualMappingCuration	DISEASE_MAPPING_COMMONS_RULES:MPR2
+HP:0008551	skos:exactMatch	MP:0000018	semapv:ManualMappingCuration	DISEASE_MAPPING_COMMONS_RULES:MPR3
+HP:0000411	skos:exactMatch	MP:0000021	semapv:ManualMappingCuration	DISEASE_MAPPING_COMMONS_RULES:MPR3
--- a/examples/schema/curation_rule-propagated.sssom.tsv
+++ b/examples/schema/curation_rule-propagated.sssom.tsv
@@ -10,9 +10,9 @@
 #mapping_provider: "https://w3id.org/sssom/core_team"
 #curation_rule:
 #  - DISEASE_MAPPING_COMMONS_RULES:MPR2
-
 #comment: This is an example file for the SSSOM for illustration only. Its contents are entirely fabricated.
 subject_id	predicate_id	object_id	mapping_justification
 HP:0009124	skos:exactMatch	MP:0000003	semapv:ManualMappingCuration
 HP:0008551	skos:exactMatch	MP:0000018	semapv:ManualMappingCuration
 HP:0000411	skos:exactMatch	MP:0000021	semapv:ManualMappingCuration
+
--- a/examples/schema/curation_rule-propagated.sssom.tsv
+++ b/examples/schema/curation_rule-propagated.sssom.tsv
@@ -4,13 +4,14 @@
 #  orcid: https://orcid.org/
 #  DISEASE_MAPPING_COMMONS_RULES: https://w3id.org/sssom/commons/disease/curation-rules/
 #mapping_set_id: https://w3id.org/sssom/commons/examples/curation_rule.sssom.tsv
-#mapping_set_description: "This example illustrates how to express that all mapppings in a mapping set have been curated according to a specific curation rule"
+#mapping_set_description: "This example illustrates how to express that all mappings in a mapping set have been curated according to a specific curation rule"
 #license: "https://creativecommons.org/publicdomain/zero/1.0/"
 #creator_id: orcid:0000-0002-7356-1779
 #mapping_provider: "https://w3id.org/sssom/core_team"
-#curation_rule: DISEASE_MAPPING_COMMONS_RULES:MPR2
+#curation_rule:
+#  DISEASE_MAPPING_COMMONS_RULES:MPR2
 #comment: This is an example file for the SSSOM for illustration only. Its contents are entirely fabricated.
 subject_id	predicate_id	object_id	mapping_justification
-HP:0009124	skos:exactMatch	MP:0000003	semapv:ManualMappingCuration	DISEASE_MAPPING_COMMONS_RULES:MPR2
-HP:0008551	skos:exactMatch	MP:0000018	semapv:ManualMappingCuration	DISEASE_MAPPING_COMMONS_RULES:MPR3
-HP:0000411	skos:exactMatch	MP:0000021	semapv:ManualMappingCuration	DISEASE_MAPPING_COMMONS_RULES:MPR3
+HP:0009124	skos:exactMatch	MP:0000003	semapv:ManualMappingCuration
+HP:0008551	skos:exactMatch	MP:0000018	semapv:ManualMappingCuration
+HP:0000411	skos:exactMatch	MP:0000021	semapv:ManualMappingCuration
--- a/examples/schema/curation_rule-propagated.sssom.tsv
+++ b/examples/schema/curation_rule-propagated.sssom.tsv
@@ -9,7 +9,8 @@
 #creator_id: orcid:0000-0002-7356-1779
 #mapping_provider: "https://w3id.org/sssom/core_team"
 #curation_rule:
-#  DISEASE_MAPPING_COMMONS_RULES:MPR2
+#  - DISEASE_MAPPING_COMMONS_RULES:MPR2
+
 #comment: This is an example file for the SSSOM for illustration only. Its contents are entirely fabricated.
 subject_id	predicate_id	object_id	mapping_justification
 HP:0009124	skos:exactMatch	MP:0000003	semapv:ManualMappingCuration
--- a/examples/schema/curation_rule-propagated.sssom.tsv
+++ b/examples/schema/curation_rule-propagated.sssom.tsv
@@ -15,4 +15,3 @@ subject_id	predicate_id	object_id	mapping_justification
 HP:0009124	skos:exactMatch	MP:0000003	semapv:ManualMappingCuration
 HP:0008551	skos:exactMatch	MP:0000018	semapv:ManualMappingCuration
 HP:0000411	skos:exactMatch	MP:0000021	semapv:ManualMappingCuration
-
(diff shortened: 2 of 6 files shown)


gouttegd (approved):
Looks good.

You may want to update the [section about propagatable slots](https://mapping-commons.github.io/sssom/spec-model/#propagation-of-mapping-set-slots) in the spec (`src/docs/spec-model.md`), to add `curation_rule_*` to the list of propagatable slots.

(Ideally that list should be somehow automatically generated from the schema; alternatively, we can just remove the list – it is just here as a convenience after all.)

gouttegd:
As for the test suite:

I initially ran into a typically obscure `'name'` error, which can be traced back to the merging of #462. Upgrading Poetry to 2.1.3 cleared the issue.

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@
 - Add `mapping_tool_id` slot to the `Mapping` and `MappingSet` classes ([issue](https://github.com/mapping-commons/sssom/issues/449)).
 - Add `record_id` slot to the `Mapping` class ([issue](https://github.com/mapping-commons/sssom/issues/359)).
 - Change all URI-typed slots to clarify that they expect _non-relative_ URIs as values ([issue](https://github.com/mapping-commons/sssom/issues/448)).
+- `curation_rule` and `curation_rule_text` are made propagatable ([issue](https://github.com/mapping-commons/sssom/issues/464)).
```
gouttegd:
```suggestion
- `curation_rule` and `curation_rule_text` have been added to the `MappingSet` class and made propagatable ([issue](https://github.com/mapping-commons/sssom/issues/464)).
```

gouttegd (dismissed):
Looks good this time.

Unrelated note: we should probably add to the PR template something like:

```
If you are proposing a change to the SSSOM metadata model, you must:

...
- [ ] update the "Model changes across versions" (in `src/docs/spec-models.md`) accordingly
```

gouttegd (dismissed):

cthoyt (changes requested):
I would like to see an explicit, _real_ example of this being used in an SSSOM document

src/docs/spec-model.md
```diff
@@ -59,7 +59,9 @@ For convenience, here is the current list of propagatable slots:
 * `subject_source_version`,
 * `subject_type`,
 * `predicate_type`,
-* `similarity_measure`.
+* `similarity_measure`,
+* `curation_rule`,
+* `curation_rule_text`.
```
ehartley: To maintain the alphabetical order of the list, I suggest adding these terms after `cardinality_scope` instead of at the end of the list.

ehartley (approved):

examples/schema/curation_rule-propagated.sssom.tsv
```diff
@@ -0,0 +1,16 @@
...
+#  orcid: https://orcid.org/
+#  DISEASE_MAPPING_COMMONS_RULES: https://w3id.org/sssom/commons/disease/curation-rules/
+#mapping_set_id: https://w3id.org/sssom/commons/examples/curation_rule.sssom.tsv
+#mapping_set_description: "This example illustrates how to express that all mapppings in a mapping set have been curated according to a specific curation rule"
+#license: "https://creativecommons.org/publicdomain/zero/1.0/"
+#creator_id: orcid:0000-0002-7356-1779
+#mapping_provider: "https://w3id.org/sssom/core_team"
+#curation_rule: DISEASE_MAPPING_COMMONS_RULES:MPR2
+#comment: This is an example file for the SSSOM for illustration only. Its contents are entirely fabricated.
+subject_id	predicate_id	object_id	mapping_justification
```
gouttegd, line 14: The header for the fifth column is missing.

But in fact I don’t think the fifth column should even be there. The point of making `curation_rule` propagatable is to allow to specify the curation rule(s) once and for all, at the level of the mapping set, if the same rules are used by all mappings.

Even if that example was correct (if the column header was not missing), it would not demonstrate the usefulness of making `curation_rule` propagatable, since the three mappings do _not_, in this example, use the same rules.

So I’d suggest removing the fifth column entirely.

gouttegd (changes requested):

examples/schema/curation_rule-propagated.sssom.tsv
```diff
@@ -0,0 +1,16 @@
...
+#  HP: http://purl.obolibrary.org/obo/HP_
+#  MP: http://purl.obolibrary.org/obo/MP_
+#  orcid: https://orcid.org/
+#  DISEASE_MAPPING_COMMONS_RULES: https://w3id.org/sssom/commons/disease/curation-rules/
+#mapping_set_id: https://w3id.org/sssom/commons/examples/curation_rule.sssom.tsv
+#mapping_set_description: "This example illustrates how to express that all mapppings in a mapping set have been curated according to a specific curation rule"
+#license: "https://creativecommons.org/publicdomain/zero/1.0/"
+#creator_id: orcid:0000-0002-7356-1779
+#mapping_provider: "https://w3id.org/sssom/core_team"
+#curation_rule: DISEASE_MAPPING_COMMONS_RULES:MPR2
```
gouttegd:
```suggestion
#curation_rule:
#  - DISEASE_MAPPING_COMMONS_RULES:MPR2

```

`curation_rule` is a multi-valued slot. Even if there is only one value, it is supposed to appear as a _list_ in the YAML metadata block.

Representing a multi-valued slot with a single value as if it was a single-valued slot is something that the spec tolerates only [as an exception](https://mapping-commons.github.io/sssom/spec-formats-tsv/#multi-valued-slots-with-a-single-value), solely for backwards compatibility with SSSOM-Py. This is discouraged and it should not be used, especially in our own examples.

examples/schema/curation_rule-propagated.sssom.tsv
matentzn: This was what I intended to do, should have run the test locally

src/docs/spec-model.md
matentzn: Thank you very much, fixed!

examples/schema/curation_rule-propagated.sssom.tsv
```diff
@@ -0,0 +1,17 @@
...
+#  MP: http://purl.obolibrary.org/obo/MP_
+#  orcid: https://orcid.org/
+#  DISEASE_MAPPING_COMMONS_RULES: https://w3id.org/sssom/commons/disease/curation-rules/
+#mapping_set_id: https://w3id.org/sssom/commons/examples/curation_rule.sssom.tsv
+#mapping_set_description: "This example illustrates how to express that all mappings in a mapping set have been curated according to a specific curation rule"
+#license: "https://creativecommons.org/publicdomain/zero/1.0/"
+#creator_id: orcid:0000-0002-7356-1779
+#mapping_provider: "https://w3id.org/sssom/core_team"
+#curation_rule:
+#  DISEASE_MAPPING_COMMONS_RULES:MPR2
```
gouttegd:
```suggestion
#  - DISEASE_MAPPING_COMMONS_RULES:MPR2

```

Otherwise it’s still a scalar value, that is simply written on two lines.

examples/schema/curation_rule-propagated.sssom.tsv
```diff
@@ -0,0 +1,17 @@
...
+#  MP: http://purl.obolibrary.org/obo/MP_
+#  orcid: https://orcid.org/
+#  DISEASE_MAPPING_COMMONS_RULES: https://w3id.org/sssom/commons/disease/curation-rules/
+#mapping_set_id: https://w3id.org/sssom/commons/examples/curation_rule.sssom.tsv
+#mapping_set_description: "This example illustrates how to express that all mappings in a mapping set have been curated according to a specific curation rule"
+#license: "https://creativecommons.org/publicdomain/zero/1.0/"
+#creator_id: orcid:0000-0002-7356-1779
+#mapping_provider: "https://w3id.org/sssom/core_team"
+#curation_rule:
+#  - DISEASE_MAPPING_COMMONS_RULES:MPR2
```
cthoyt, line 12: can you please use an example of a curation rule that can be resolved when expanded to a URI?

I think for most SSSOM curators/developers/users, it isn't clear what a curation rule is nor why you should use one, so it would be important that the example uses a curation rule that has some real context attached to it

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@
 - Add `mapping_tool_id` slot to the `Mapping` and `MappingSet` classes ([issue](https://github.com/mapping-commons/sssom/issues/449)).
 - Add `record_id` slot to the `Mapping` class ([issue](https://github.com/mapping-commons/sssom/issues/359)).
 - Change all URI-typed slots to clarify that they expect _non-relative_ URIs as values ([issue](https://github.com/mapping-commons/sssom/issues/448)).
+- Add `curation_rule` and `curation_rule_text` to the `MappingSet` class and made propagatable ([issue](https://github.com/mapping-commons/sssom/issues/464)).
```
cthoyt, line 15: can you give an example that demonstrates what a curation rule text is, and it being used in a propgatable way?

examples/schema/curation_rule-propagated.sssom.tsv
gouttegd: @cthoyt I’d say: outside of the scope of this PR.

If we were adding completely new slots, I could agree, but here, both `curation_rule` and `curation_rule_text` have existed for a while, this PR is solely about making them propagatable.

If you think that some existing slots do not have enough meaningful examples, I’d rather we open another issue and tackle that problem separately.

nichtich: We use some curation rules in our mapping projects, see https://github.com/gbv/jskos/issues/153 for planned reference in JSKOS field `guidelines` that maps to SSSOM `curation_rule`. All instances are URLs of PDF or HTML documents explaining how to curate mappings between specific vocabularies.

matentzn: Thanks @nichtich and @cthoyt -> @gouttegd and I moved this to an issue and will address this as a separate documenation effort: https://github.com/mapping-commons/sssom/issues/479

CHANGELOG.md
matentzn: We will address this separately: https://github.com/mapping-commons/sssom/issues/479
Propagation is documented in detail here: https://mapping-commons.github.io/sssom/spec-formats-tsv/#propagation (basically, all the values in the mapping set apply to all individual mappings in the set if the slit is "propagatable").

cthoyt: I think you misunderstand. You're requesting to modify the spec in this PR without providing an example of how that change might look in practice

Even if we throw out the idea that the example should be meaningful, I insist that you at least put _something_ that shows `curation_rule_text` being used in the metadata at the top of a SSSOM document

examples/schema/curation_rule-propagated.sssom.tsv
cthoyt: I'm not happy to see an example that's not meaningful, but I concede that my request is getting in the way, so I'd be happy to see the proposed example improved in a follow-up PR as an alternative

matentzn:
@cthoyt would this example satisfy you?

```
#curie_map:
#  HP: http://purl.obolibrary.org/obo/HP_
#  MP: http://purl.obolibrary.org/obo/MP_
#  orcid: https://orcid.org/
#mapping_set_id: https://w3id.org/sssom/commons/examples/curation_rule.sssom.tsv
#mapping_set_description: "This example illustrates how to express that all mappings in a mapping set have been curated according to a specific curation rule"
#license: "https://creativecommons.org/publicdomain/zero/1.0/"
#creator_id: orcid:0000-0002-7356-1779
#mapping_provider: "https://w3id.org/sssom/core_team"
#curation_rule_text:
#  - Human and mouse phenotypes that inhere in homologous structures and exhibit the same phenotypic quality are considered exact matches.
#comment: This is an example file for the SSSOM for illustration only. Its contents are entirely fabricated.
subject_id	predicate_id	object_id	mapping_justification
HP:0009124	skos:exactMatch	MP:0000003	semapv:ManualMappingCuration
HP:0008551	skos:exactMatch	MP:0000018	semapv:ManualMappingCuration
HP:0000411	skos:exactMatch	MP:0000021	semapv:ManualMappingCuration
```

examples/schema/curation_rule_text-propagated.sssom.tsv
```diff
@@ -0,0 +1,16 @@
+#curie_map:
+#  HP: http://purl.obolibrary.org/obo/HP_
+#  MP: http://purl.obolibrary.org/obo/MP_
+#  orcid: https://orcid.org/
+#mapping_set_id: https://w3id.org/sssom/commons/examples/curation_rule.sssom.tsv
+#mapping_set_description: "This example illustrates how to express that all mappings in a mapping set have been curated according to a specific curation rule"
+#license: "https://creativecommons.org/publicdomain/zero/1.0/"
+#creator_id: orcid:0000-0002-7356-1779
```
gouttegd, line 8: Once again I must request that our examples do _not_ perpetuate a syntax that we only accept in the spec for backwards compatibility with SSSOM-Py.

`creator_id` is a multi-valued slot, it expects a _list_, not a scalar.

```suggestion
#creator_id:
#  - orcid:0000-0002-7356-1779
```

gouttegd: Oh, merged already. OK, I guess we don’t give a damn.

matentzn: Sorry about that: https://github.com/mapping-commons/sssom/pull/481

Issue #467, opened 2025-07-12
The meaning of `mapping_cardinality` must be clarified


gouttegd:
The `mapping_cardinality` slot is currently “defined” as follows:

> A string indicating whether this mapping is from a 1:1 (the subject_id maps to a single object_id), 1:n (the subject maps to more than one object_id), n:1, 1:0, 0:1 or n:n group. Note that this is a convenience field that should be derivable from the mapping set.

But if one goes to have a look at [the issue](https://github.com/mapping-commons/sssom/issues/53) in which this slot was proposed, one can see that the initial definition was (emphasis mine):

> given a mapping <s,p,o>
>
> * `1:n`: s is mapped to more than o **with relation p**
> * `1:1`: s is mapped to exactly 1 o **with relation p**
> * `n:1`: more than 1 s is mapped to the same o with **relation p**
> * `1:0`: o is "sssom:NoMatch"
> * `0:1`: s is "sssom:NoMatch"

“**with relation p**”: this precision did not make it to the actual definition in the model, and this changes the meaning of the “cardinality” quite significantly.

Under the original definition (the one proposed in the ticket), cardinality takes into account not only the subject and the object, but also the _predicate_ between the two. In the final definition (the one in the actual LinkML schema), only the subject and the object matters, _not_ the predicate.

So considering the following mappings:

```
subject_id       predicate_it       object_id
UBERON:0000011   skos:broadMatch    VHOG:0000755
UBERON:0000011   skos:narrowMatch   EHDAA:4655
UBERON:0000011   skos:narrowMatch   NCIT:C12764
UBERON:0000011   skos:narrowMatch   SCTID:362496006
```

Under the original definition, the last three have a `1:n` cardinality (same subject mapped to more than one object with the same relation `skos:narrowMatch`), but the first one has a `1:1` cardinality (it has the same subject than the other three, but a different predicate, and there is no other mapping with the same subject, the same predicate, but a different object).

Under the final definition, the first mapping has a `1:n` cardinality like the other three, because the fact that it has a different predicate does not matter – all that matters is that there are other mappings with the same subject but different objects.

So, which definition is correct then?

gouttegd:
And if the original definition is the correct one, and the predicate is supposed to matter, then how does `predicate_modifier` interact with cardinality?

[...]

Issue #53, opened 2021-03-07
New metadata element: mapping_cardinality


matentzn:
Proposal: add a new field `mapping_cardinality` which can take the following values:
1:n, 1:1, n:1, 1:0, 0:1

These are in principle derivable from the mapping set, but turn out to be super useful for querying and analysis to begin with.

given a mapping <s,p,o>
- `1:n`: s is mapped to more than o with relation p
- `1:1`: s is mapped to exactly 1 o with relation p
- `n:1`: more than 1 s is mapped to the same o with relation p
- `1:0`: o is "sssom:NoMatch"
- `0:1`: s is "sssom:NoMatch"

matentzn:
From slack:

```
enums:
  status_enum:
    permissible_values:
      development: 
      release:
slots:
  status:
     range: status_enum
     required: false
```

matentzn:
There are some issue atm with the enums in linkml, but this is handled [elsewhere](https://github.com/mapping-commons/sssom-py/pull/52). The property is added.

PR #476, opened 2025-07-31
Add `cardinality_scope` slot.

gouttegd:
Resolves [#467]

- [x] make sure any new slot is annotated with the appropriate `added_in` annotation

This PR adds a new mapping record slot called `cardinality_scope`, intended to precise the meaning of the value of the `mapping_cardinality` slot.

If empty, it means that the cardinality is relative to the entire set (which corresponds to the current definition of `mapping_cardinality`, and therefore ensures backwards compatibility). If non-empty, it is expected to be a list of slots that are used to define the subset (the "scope") of mapping records in which the cardinality has been computed.

For example, a cardinality scope of (`predicate_id`, `object_source`) means that the cardinality is computed over the subset of mapping records that have the same predicate and whose objects come from the same source.

The new slot is propagatable, because it is assumed that in most mapping sets, the cardinality for all records will be computed in the same way (though there's no obligation for that), and so all records will share the same scope.

### DIFF
--- a/examples/schema/cardinality-scope-predicate+object_source.sssom.tsv
+++ b/examples/schema/cardinality-scope-predicate+object_source.sssom.tsv
@@ -1,0 +1,17 @@
+#curie_map:
+#  COMENT: https://example.com/entities/
+#  NETENT: https://example.net/entities/
+#  ORGENT: https://example.org/entities/
+#  SRC: https://example.org/sources/
+#mapping_set_id: https://example.org/sets/cardinality-scope-predicate+object_source
+#license: https://creativecommons.org/licenses/by/4.0/
+#cardinality_scope:
+#  - predicate_id
+#  - object_source
+subject_id	subject_label	predicate_id	object_id	object_label	mapping_justification	object_source	mapping_cardinality
+ORGENT:0001	alice	skos:closeMatch	COMENT:0011	alpha	semapv:ManualMappingCuration	SRC:com	1:1
+ORGENT:0001	alice	skos:closeMatch	NETENT:0111	alpha	semapv:ManualMappingCuration	SRC:net	1:1
+ORGENT:0002	bob	skos:closeMatch	COMENT:0012	beta	semapv:ManualMappingCuration	SRC:com	1:1
+ORGENT:0002	bob	skos:closeMatch	NETENT:0112	bravo	semapv:ManualMappingCuration	SRC:net	1:1
+ORGENT:0007	gavin	skos:closeMatch	NETENT:0117	golf	semapv:ManualMappingCuration	SRC:net	1:1
+ORGENT:0007	gavin	skos:exactMatch	COMENT:0013	gamma	semapv:ManualMappingCuration	SRC:com	1:1
--- a/src/docs/spec-model.md
+++ b/src/docs/spec-model.md
@@ -128,6 +128,53 @@ The meaning of the NOT predicate modifier in a mapping that refers to `sssom:NoT
 When computing cardinality values (to fill the `mapping_cardinality` slot), mappings that refer to `sssom:NoTermFound` MUST be ignored.
 
 
+## Mapping cardinality and cardinality scope
+
+The `mapping_cardinality` slot is somewhat special in that its value is
+only meaningful within a given context, or “scope”: a mapping record in
+itself does not have any cardinality – it only has one when it is part
+of a larger set of records.
+
+Consider the following three records (set metadata, and in particular
+prefix declarations, have been omitted for brevity):
+
+| `subject_id`   | `predicate_id`   | `object_id`  | `object_source` |
+| -------------- | ---------------- | ------------ | --------------- |
+| UBERON:0000011 | skos:broadMatch  | VHOG:0000755 | obo:VHOG        |
+| UBERON:0000011 | skos:narrowMatch | EHDAA:4655   | obo:EHDAA       |
+| UBERON:0000011 | skos:narrowMatch | NCIT:C12764  | obo:NCIT        |
+
+Within that particular set, all three records have a cardinality of
+`1:n` (one subject, UBERON:0000011, mapped to many objects).
+
+But cardinality can also be computed on smaller subsets. For example:
+
+* if we are only interested in records that have the same predicate,
+  then the first record has a cardinality of `1:1` (UBERON:0000011 is
+  mapped to only one object through a `skos:broadMatch` predicate),
+  while the other two still have a cardinality of `1:n` (UBERON:0000011
+  is mapped to two different objects through a `skos:narrowMatch`
+  predicate);
+* if we are only interested in records where the objects are from the
+  same source, then all three records have a cardinality of `1:1`
+  (UBERON:0000011 is mapped to only one object in each of the three
+  vocabularies VHOG, EHDAA, and NCIT).
+
+It is left to users and downstream applications of SSSOM to decide which
+type of cardinality (relative to the entire set or relative to any of
+the many possible subsets) will be the most useful to them. The
+`cardinality_scope` slot is intended to allow them to specify which
+cardinality they use.
+
+When computing cardinality values:
+
+* if the cardinality is computed on the entire set, the
+  `cardinality_scope` slot MUST be left empty (or absent);
+* if the cardinality is computed on a subset, the `cardinality_scope`
+  slot MUST be filled with the list of slots that are used to define the
+  subset.
+
+
 ## Non-standard slots
 
 <a id="non-standard-slots"></a>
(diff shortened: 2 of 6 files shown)


examples/schema/cardinality-scope-predicate+object_source.sssom.tsv
matentzn, line 1: is the plus sign a good idea in a file path?

src/docs/spec-model.md
```diff
@@ -128,6 +128,53 @@ The meaning of the NOT predicate modifier in a mapping that refers to `sssom:NoT
...
+It is left to users and downstream applications of SSSOM to decide which
+type of cardinality (relative to the entire set or relative to any of
+the many possible subsets) will be the most useful to them. The
+`cardinality_scope` slot is intended to allow them to specify which
+cardinality they use.
+
+When computing cardinality values:
+
+* if the cardinality is computed on the entire set, the
+  `cardinality_scope` slot MUST be left empty (or absent);
```
matentzn, line 173: > computed on the entire set

What exactly does this mean? Does this include _all_ sssom metadata fields? That seems not what was originally intended - imagine this case:

```
| `subject_id`   | `predicate_id`   | `object_id`  | `object_source` | comment |
| -------------- | ---------------- | ------------ | --------------- | ------- |
| UBERON:0000011 | skos:narrowMatch | VHOG:0000755   | obo:VHOG       | bar |
| UBERON:0000011 | skos:narrowMatch | VHOG:0000755  | obo:VHOG        | foo |
```

The intention was hardly intended to say that these two are 1:n mappings, right?

((((I just though 15 minutes about what we did mean orginally and I keep hitting one wall or another - hence very welcome to clarify this in this PR, thanks)))

... Still not sure what we meant originally. I think historically, its something like this:

1. We (wrongly) assumed that a mapping set is scoped by `object_source` and `subject_source` (or to be precise, their versions, but lets forget about that for now). Sometimes the scope is refined by "predicate_id" (mappings  between Uberon and FBBT / exact mappings between Uberon and FBBT). 
2. We (wrongly) assumed that a mapping set is _complete_ with respect to that scope (i.e. **all** exact mappings between Uberon and FBBT).
3. Under the two above (wrong) assumptions, we then defined cardinality based solely on subject_id, object_id.

So, since all the original assumptions are mostly false now (especially assumption 2), the original _idea behind the design_ (not necessarily its flawed implementation) does not hold anymore.

gouttegd: > The intention was hardly intended to say that these two are 1:n mappings, right?

Of course not. Those are `1:1` mappings. When computing cardinality (and regardless of the “scope” in which you do that), you always look at the number of _different_ subjects and objects.

* For every given subject, how many _different_ objects are mapped to it? If only 1, the right side of the cardinality is 1; if more than 1, it’s _n_.
* For every given object, how many _different_ subjects are mapped to it? If only 1, the left side of the cardinality is 1; if more than 1, it’s _n_.

> We (wrongly) assumed that a mapping set is scoped by object_source and subject_source 

Speak for yourself. AFAIK that assumption _never_ appeared anywhere, and I never made it.

> Sometimes the scope is refined by "predicate_id"

This is what appeared in your original #53 issue, but as I mentioned in #467 this never made it to the description in the schema. I never made that assumption as well (I implemented cardinality computation in SSSOM-Java based solely on the description available in the model).

> So, since all the original assumptions are mostly false now (especially assumption 2), the original idea behind the design (not necessarily its flawed implementation) does not hold anymore.

From my point of view, the one false assumption is the one that there is _one_ definition of cardinality, or if you prefer, one _absolute_ cardinality. That is clearly not the case. This PR intends to make clear that cardinality is _relative_, with the `cardinality_scope` slot providing the information about what it is relative to.

gouttegd: > > computed on the entire set
> 
> What exactly does this mean?

That you take into account _all_ mapping records in the set, as opposed to the case where you only take into account a _subset_ of the records – for example, the subset of mapping records with the same object source than the record for which you want to compute the cardinality.

examples/schema/cardinality-scope-predicate+object_source.sssom.tsv
gouttegd: I am not aware of any filesystem where a `+` character is a problem, not even on Windows.

src/docs/spec-model.md
matentzn: > Speak for yourself. AFAIK that assumption never appeared anywhere, and I never made it.

No I meant, in 2021 when the mapping_cardinality slot was first discussed. Anyways, not important now, just trying to jog my memory.

> From my point of view, the one false assumption is the one that there is one definition of cardinality, or if you prefer, one absolute cardinality.

Yes totally.

Sorry, maybe my example was wrong:

```
| `subject_id`   | `predicate_id`   | `object_id`  | `object_source` | comment |
| -------------- | ---------------- | ------------ | --------------- | ------- |
| UBERON:0000011 | skos:narrowMatch | VHOG:0000755   | obo:VHOG | bar |
| UBERON:0000011 | skos:narrowMatch | VHOG:0000999  | obo:VHOG | foo |
```

These should be two 1:n mappings..

And maybe my brain is not quite working well.

examples/schema/cardinality-scope-predicate+object_source.sssom.tsv
matentzn: No files system, I was just thinking URLs, but of course these are correctly encoded.

src/docs/spec-model.md
gouttegd: > These should be two 1:n mappings..

Yes, and they are:

* Consider the first record;
    * the subject is UBERON:0000011; how many different objects are mapped to UBERON:0000011? Two (VHOG:0000755 and VHOG:0000999), so the right side of the cardinality is `:n`;
    * the object is VHOG:0000755; how many different subjects are mapped to VHOG:0000755? Only one (UBERON:0000011), so the left side of the cardinality is `1:`;
    * so the cardinality for that record is `1:n`.
* Consider the second record:
    * the subject is UBERON:0000011; how many different objects are mapped to UBERON:0000011? Two (we already checked that above), so the right side of the cardinality is `:n`;
    * the object is VHOG:0000999; how many different subjects are mapped to VHOG:0000999? Only one (UBERON:0000011), so the left side of the cardinality is `1:`;
    * so the cardinality for that record is `1:n`.

matentzn: Moving to slack in the deepest secret corner to avoid embarassing myself in front of the whole world.

gouttegd: I am pretty sure that the whole world has better (or worse… 😞 ) things to do than to lurk on the SSSOM repo to make fun of cardinality computation problems.

matentzn: Alright, got it now. What you call `subset` is more like an aggregator/group_by slot: if nothing is selected (the default) subject_id x object_id combinations are all that matters for cardinality computation (this is what tripped me up: the “whole mapping set” evoked “all of the mapping slots”). If you select a “subset” what you do is you basically introduce some kind of group-by logic, saying “given the same predicate_id” or given the same “predicate_id, object_source” combination.

I am now fully onboard with the PR as is.

src/sssom_schema/schema/sssom_schema.yaml
```diff
@@ -584,6 +584,38 @@ slots:
...
+      propagated: true
+      added_in: "1.1"
+    examples:
+      - value:
+          - "predicate_id"
+        description: Indicates that mapping_cardinality is computed relatively to all
+          mappings that have the same predicate.
+      - value:
+          - "predicate_id"
+          - "object_source"
```
matentzn, line 612: I do not myself know if this syntax is valid, or you need to use something more like 

```suggestion
          - ["object_source", "predicate_id"]
```

But since the qc is not complaining, lets let it stand.

matentzn (dismissed):

src/sssom_schema/schema/sssom_schema.yaml
gouttegd: Well, it’s _either_ block syntax:

```yaml
value:
  - "predicate_id"
  - "object_source"
```

_or_ inline syntax:

```yaml
value: [ "predicate_id", "object_source" ]
```

but not both.

This:

```yaml
value:
  - [ "predicate_id", "object_source" ]
```

is valid but completely different: it defines the value as being a list containing a single element which is itself a list of two elements (instead of simply a list of two elements).

src/docs/spec-model.md
ehartley, line 1: Since it is propagatable, `cardinality_scope` should be added to the list of propagatable slots in the "Propagation of mapping set slots" section.

ehartley (dismissed):
All the changes look good. My only suggestion is to update the list of propagatable slots in the spec-model.md file.

src/docs/spec-model.md
gouttegd: Absolutely, my bad.

ehartley (approved):

matentzn (approved):

PR #480, opened 2025-08-11
Add `0:0` special cardinality value.

### LINKED ISSUE #477: Deprecate the `1:0` and `0:1` values in the mapping_cardinality_enum
I think the `1:0` and `0:1` values in the `mapping_cardinality_enum` currently serve no purpose and should be deprecated.

The meaning of those values is not explained in the model, but according to the [original issue](https://github.com/mapping-commons/sssom/issues/53) they were intended as a way to express the fact that a term in a source vocabulary had no corresponding match in a target vocabulary:

* `1:0`: o is "sssom:NoMatch"
* `0:1`: s is "sssom:NoMatch"

If that is indeed their intended meaning, then it is redundant with the `sssom:NoTermFound` special identifier.

This means that the absence that the absence of a match on the object side can be represented in two different ways:

* either by setting `mapping_cardinality` to `1:0`,
* or by setting the `object_id` to `sssom:NoTermFound`.

I don’t think it is a good idea to have two different mechanisms to represent the same thing. And since `sssom:NoTermFound`, contrary to the `1:0` cardinality, _is_ [documented in SSSOM 1.0](https://mapping-commons.github.io/sssom/spec-model/#representing-unmapped-entities), this is the mechanism we should keep – and therefore we should deprecate the `1:0` / `0:1` cardinality values.

“Deprecating” here meaning that the values would stay around (to avoid breaking existing sets), but implementations would be instructed _not_ to emit them, and to ignore them when they are found in existing sets.

gouttegd:
Resolves [#477]

This PR adds a new cardinality value of `0:0`, to be used for mapping records where both the `subject_id` and the `object_id` are `sssom:NoTermFound`. Such records are allowed and may be used to represent the fact that there is no overlap between a subject vocabulary and an object vocabulary.

The spec is updated to explain how mappings involved `sssom:NoTermFound` should be handled when computing cardinality values.

We also try to explain a bit more what the different cardinality values mean exactly, and provide two examples:

* one that illustrates all possible cardinality values between real mappings (1:1, 1:n, n:1, n:n);
* one that illustrates the special cases where the subject_id and/or the object_id is sssom:NoTermFound.

closes #477

### DIFF
--- a/src/sssom_schema/schema/sssom_schema.yaml
+++ b/src/sssom_schema/schema/sssom_schema.yaml
@@ -929,6 +929,18 @@ classes:
     slot_usage:
       license:
         required: true
+      similarity_measure:
+        instantiates: sssom:Versionable
+        annotations:
+          added_in: "1.1"
+      curation_rule:
+        instantiates: sssom:Versionable
+        annotations:
+          added_in: "1.1"
+      curation_rule_text:
+        instantiates: sssom:Versionable
+        annotations:
+          added_in: "1.1"
     slots:
     - sssom_version
     - curie_map


examples/schema/cardinality-with-unmapped-entities.sssom.tsv
```diff
@@ -0,0 +1,11 @@
...
+#  OBJ: https://example.org/object/
+#  SRC: https://example.org/sources/
+#  SUBJ: https://example.org/subject/
+#mapping_set_id: https://example.org/sets/cardinality-with-unmapped-entities
+#license: https://creativecommons.org/licenses/by/4.0/
+subject_id	predicate_id	object_id	mapping_justification	subject_source	object_source	mapping_cardinality	comment
+SUBJ:0001	skos:exactMatch	sssom:NoTermFound	semapv:ManualMappingCuration	SRC:A	SRC:B	1:0	S1 in vocabulary A has no exact match in vocabulary B
+SUBJ:0001	skos:closeMatch	OBJ:0001	semapv:ManualMappingCuration	SRC:A	SRC:B	1:1	S1 mapped only to O1, O1 mapped only to S1 -- the record involving sssom:NoTermFound does not count, as it is an absence of match rather than an actual mapping
+sssom:NoTermFound	skos:exactMatch	OBJ:0002	semapv:ManualMappingCuration	SRC:C	SRC:D	0:1	O2 in vocabulary D has no exact match in vocabulary C
+sssom:NoTermFound	skos:exactMatch	sssom:NoTermFound	semapv:ManualMappingCuration	SRC:E	SRC:F	0:0	No exact match between any term ffrom vocabulary E and any term for vocabulary F (in other words, the two vocabularies are completely disjoint, at least as far as exact matches are considered)
```
matentzn:
```suggestion
sssom:NoTermFound	skos:exactMatch	sssom:NoTermFound	semapv:ManualMappingCuration	SRC:E	SRC:F	0:0	No exact match between any term from vocabulary E and any term for vocabulary F (in other words, the two vocabularies are completely disjoint, at least as far as exact matches are considered)
```

src/docs/spec-model.md
```diff
@@ -126,7 +126,7 @@ The `sssom:NoTermFound` value MUST NOT be used in any other slot than `subject_i
 
 The meaning of the NOT predicate modifier in a mapping that refers to `sssom:NoTermFound` is unspecified.
 
-When computing cardinality values (to fill the `mapping_cardinality` slot), mappings that refer to `sssom:NoTermFound` MUST be ignored.
+When computing cardinality values (to fill the `mapping_cardinality` slot): (1) a mapping record with a `object_id` (respectively `subject_id`) of `sssom:NoTermFound` MUST be assigned a cardinality value of `1:0` (respectively `0:1`), regardless of any other record; (2) a mapping record with both the `subject_id` and the `object_id` set to `sssom:NoTermFound` MUST be assigned a cardinality value of `0:0`, regardless of any other record; (3) such records MUST be ignored when computing the cardinality of other records.
```
matentzn, line 131:
```suggestion
When computing cardinality values (to fill the `mapping_cardinality` slot): (1) a mapping record with a `object_id` (respectively `subject_id`) of `sssom:NoTermFound` (and `subject_id` (respectively `object_id`) _not_ of `sssom:NoTermFound`) MUST be assigned a cardinality value of `1:0` (respectively `0:1`), regardless of any other record; (2) a mapping record with both the `subject_id` and the `object_id` set to `sssom:NoTermFound` MUST be assigned a cardinality value of `0:0`, regardless of any other record; (3) such records MUST be ignored when computing the cardinality of other records.
```

src/docs/spec-model.md
```diff
@@ -126,7 +126,7 @@ The `sssom:NoTermFound` value MUST NOT be used in any other slot than `subject_i
...
```
matentzn, line 132:
```suggestion
* For the purpose of cardinality computation, `subject_id`s which correspond to complex expressions such as `http://my.org/schema/0001/(disease1:'MONDO:0005148',disease2:'MONDO:0000960',excluded:'HP:0100758')` (see [here](https://github.com/monarch-initiative/uri-expression-language)) SHOULD be treated as if they are single-valued.
```

Just a weak suggestion as it is a question that immediately popped into my mind.

examples/schema/cardinality-with-unmapped-entities.sssom.tsv
gouttegd: Sigh... I pushed two commits to fix typos and I _still_ missed one. 😅

src/docs/spec-model.md
gouttegd: This is not needed (and it makes the rule less easy to read): the second rule takes care of records where both sides are `sssom:NoTermFound`, so it is already clear that the first rule can only apply to mappings where only one side is `sssom:NoTermFound`.

src/docs/spec-model.md
gouttegd: No need. “Complex” identifiers are opaque as far as SSSOM is concerned. Even if a complex identifier were to involve 9 different entities in a complex expression, it would still always count for a single subject.

src/docs/spec-model.md
matentzn: Sounds good, just wanted to signal that the frst rule was not self contained, but I agree, my addition makes it harder to read.

src/docs/spec-model.md
matentzn: Of course, I am merely pointing out that this would be something I would ask as a user - I guess I can see that it does not belong in the spec part for cardinality.

matentzn (dismissed):
Looks great! Thanks a ton!

src/sssom_schema/schema/sssom_schema.yaml
```diff
@@ -76,12 +76,30 @@ enums:
...
+      "n:n":
+        description: Indicates the mapping record is about a many-to-many mapping, that
+          is, the subject is mapped to several different objects and the object is
+          mapped to several different subjects.
+      "1:0":
+        description: Indicates that the subject has no match in the object vocabulary.
+          This value MUST only be used when the object_id is sssom:NoTermFound.
+      "0:1":
+        description: Indicates that the object has no match in the subject vocabulary.
+          This  value MUST only be used when the subject_id is sssom:NoTermFound.
```
ehartley:
```suggestion
          This value MUST only be used when the subject_id is sssom:NoTermFound.
```
Minor typo - delete extra space.

ehartley (dismissed):
There is one minor typo, but otherwise everything looks good!

ehartley (approved):

PR #483, opened 2025-08-14
Add test for added_in annotation.

gouttegd:

This PR adds a test that any new slot must

(1) carry a `added_in` annotation that
(2) is set to the current version of the spec.

The test works by preparing a fixed compiled list of all the “non-new” slots (all slots present in the 1.0 release) for both the `Mapping` and `MappingSet` classes. It then loads the current schema and iterates over all slots for those two classes, checking that every slot that is not a pre-existing slot is annotated as expected.

Of note, the test currently fails (as it should), because the `curation_rule` and `curation_rule_text` slots are _new_ on the `MappingSet` class but are currently not annotated as such (this is what #482 is fixing).


gouttegd:
This is exactly the failure I expect for this test, until #482 has been merged:

```
FAIL: test_new_slots (__main__.TestAddedInAnnotationTestCase)
----------------------------------------------------------------------
Traceback (most recent call last):
  File "/home/runner/work/sssom/sssom/tests/test_added_in_annotations.py", line 127, in test_new_slots
    self.assertListEqual([], unannotated)
AssertionError: Lists differ: [] != ['Slot curation_rule in class mapping set [78 chars]ted']

Second list contains 2 additional elements.
First extra element 0:
'Slot curation_rule in class mapping set is not annotated'

- []
+ ['Slot curation_rule in class mapping set is not annotated',
+  'Slot curation_rule_text in class mapping set is not annotated']

----------------------------------------------------------------------
Ran 1 test in 0.052s

FAILED (failures=1)
make: *** [Makefile:56: test] Error 1
```

tests/test_added_in_annotations.py
matentzn, line 1: This is SUPER AWESOME! Thank you! And I dont know what I was thinking previously about the design!

matentzn:
@cthoyt can you indicate with a thumbs up or down if you would like to review, if not, I will find someone else! Thank you! :)

cthoyt (approved):

matentzn:
Thank you @gouttegd! Much appreciated!

PR #494, opened 2025-10-06
Update PyPI publish workflow for prereleases

### LINKED ISSUE #490: Make “development releases“ of `sssom`
The latest publicly available release of `sssom` is the 1.0.0 release, published last year.

Several new slots and enum values have been added since then, but they cannot be used by `sssom-py` because `sssom-py` has to import the `sssom` module, and obviously it can only import a _published_ version.

In effect, this means that we cannot make any work on `sssom-py` that depends on anything added to the schema after 1.0.0 – including any work intended to make `sssom-py` compliant with the future 1.1 version of the spec ([example here](https://github.com/mapping-commons/sssom-py/pull/604))! We would have to wait for SSSOM 1.1 to be published, and only _then_ we can make `sssom-py` ready for it. This is, well, not great.

Keeping in mind that the **version of the SSSOM specification** is not necessarily the same thing as the **version of the `sssom` Python module**, we should at some point make a new `sssom` release in which the new slots and enum values of the future version 1.1 of the specification are already available, so that `sssom-py` developers can work with them – and we should not wait until we believe the 1.1 version is ready for prime time.

matentzn:
Workflow now triggers on published and prereleased events, sets PEP 440-compliant version for prereleases, and adds 'skip-existing' to PyPI publish step. Improves handling of pre-release publishing and artifact management.

Resolves #490 

### DIFF
--- a/.github/workflows/pypi-publish.yaml
+++ b/.github/workflows/pypi-publish.yaml
@@ -27,8 +27,9 @@ jobs:
         pipx inject poetry poetry-dynamic-versioning
 
 
-    - name: Build sdist & wheel
-      run: poetry build
+    - name: Build source and wheel archives
+      run: |
+        poetry build
 
     - name: Store built distribution
       uses: actions/upload-artifact@v4.6.2
--- a/.github/workflows/pypi-publish.yaml
+++ b/.github/workflows/pypi-publish.yaml
@@ -26,17 +26,6 @@ jobs:
         pipx install poetry
         pipx inject poetry poetry-dynamic-versioning
 
-    - name: Set version for pre-release (PEP 440 compliant)
-      if: github.event.release.prerelease == true
-      shell: bash
-      run: |
-        BASE_VERSION="$(poetry version -s | sed 's/+.*//')"  # strip any local part
-        DATE="$(date -u +%Y%m%d)"
-        RUN="${{ github.run_number }}"
-        SHA="$(git rev-parse --short HEAD)"
-        PRERELEASE_VERSION="${BASE_VERSION}.dev${DATE}${RUN}+g${SHA}"
-        echo "Setting version to ${PRERELEASE_VERSION}"
-        poetry version "${PRERELEASE_VERSION}"
 
     - name: Build sdist & wheel
       run: poetry build
--- a/.github/workflows/pypi-publish.yaml
+++ b/.github/workflows/pypi-publish.yaml
@@ -3,8 +3,8 @@ name: Publish Python Package
 
 on:
   release:
-    # Trigger the workflow only for real releases but not for draft releases
-    types: [created]
+    # Run when a release is published or a prerelease is published (not on draft creation)
+    types: [published, prereleased]
 
 jobs:
   build:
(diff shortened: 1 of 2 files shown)


.github/workflows/pypi-publish.yaml
```diff
@@ -54,7 +65,7 @@ jobs:
           path: dist
 
       - name: Publish package 📦 to PyPI
-        if: github.event_name == 'release'
         uses: pypa/gh-action-pypi-publish@v1.12.4
         with:
-          verbose: true
\ No newline at end of file
+          verbose: true
+          skip-existing: true
```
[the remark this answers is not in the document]
matentzn: The action is only triggered when a release is created so this is redundant.

.github/workflows/pypi-publish.yaml
```diff
@@ -26,9 +26,20 @@ jobs:
...
+      shell: bash
       run: |
-        poetry build
+        BASE_VERSION="$(poetry version -s | sed 's/+.*//')"  # strip any local part
+        DATE="$(date -u +%Y%m%d)"
+        RUN="${{ github.run_number }}"
+        SHA="$(git rev-parse --short HEAD)"
+        PRERELEASE_VERSION="${BASE_VERSION}.dev${DATE}${RUN}+g${SHA}"
+        echo "Setting version to ${PRERELEASE_VERSION}"
+        poetry version "${PRERELEASE_VERSION}"
```
gouttegd: 1. Do you _really_ want the version number of a pre-release to look something like this: `1.1.0.dev20251006.1.c1dd790` ??? Why?

2. Would such a version number really be “PEP440-compliant”? Maybe it would be, but I see nothing in [said PEP](https://peps.python.org/pep-0440/) to suggest that it would be the case. On the contrary, it explicitly says:

> The canonical public version identifiers MUST comply with the following scheme:
>
> ```
> [N!]N(.N)*[{a|b|rc}N][.postN][.devN]
> ```

with nothing to suggest that the _N_ in `devN` can be anything more than a simple integer.

3. I don’t understand why any of this would be necessary. We are already using Poetry’s dynamic versioning plugin to automatically set the version number at build time based on Git tags.

So all we should have to do to publish a pre-release on PyPI with a PEP440-compliant version number is to create the pre-release on GitHub with a suitable tag.

For example, want to publish a `1.1.0a1` release?  Create a release with a `v1.1.0a1` tag. When running `poetry build` below, the dynamic versioning plugin will automatically set the version number to `1.1.0a1`, and that’s what will end up on PyPI.

Why should we need anything more complicated than that?

matentzn: 🤦 

I am getting old. I did not think of this and of course you are 100% right. Can you check again?

gouttegd (approved):
Looks good to me.

You _might_ want to add some notes somewhere that it is the responsibility of whoever wants to do a ”pre-release” to remember to create the release tag accordingly (i.e. a release tag that includes a `aN`, `bN, `.devN` suffix, depending on the desired type of pre-release).

Issue #429, opened 2025-04-06
Misleading specification of multi-valued slots in SSSOM/TSV


nichtich:
From <https://mapping-commons.github.io/sssom/spec-formats-tsv/>:

> Multi-valued slots MUST be serialised as a list of values separated by `|` characters.
>
> **Quoting**
> 
> Within the mappings block, the following quoting rules, adapted from [RFC 4180](https://datatracker.ietf.org/doc/html/rfc4180), apply:
>
> 1.  Any value MAY be enclosed in double quotes (`"`).

Here the word "value" is used in two distinct meanings. Given a multi-value slot, e.g. `creator_label` with two values *alice* and *bob*:

1. Both `alice` and `bob` are "values"
2. The serialized, unquoted list `alice|bob` is a "value"  

The specification should not use same words for different things. I recommend to remove the whole paragraph on quoting and refer to RFC 4180 instead (with a note that COMMA is replaced by HTAB to specify TSV).

And a note should be added that the character `|` cannot be expressed in SSSOM/TSV for repeatable fields. As neither Uri or CURIE allow `|`, this is only relevant to: [match_string](https://mapping-commons.github.io/sssom/match_string/), [curation_rule_text](https://mapping-commons.github.io/sssom/curation_rule_text/), [reviewer_label](https://mapping-commons.github.io/sssom/reviewer_label/), [author_label](https://mapping-commons.github.io/sssom/author_label/), [creator_label](https://mapping-commons.github.io/sssom/creator_label/). Maybe the data model could be adjusted to use data type "String-without |" instead of arbitrary strings.

gouttegd:
> 1. Both alice and bob are "values">
> 2. The serialized, unquoted list alice|bob is a "value"
>
> The specification should not use same words for different things.

I agree, this is needlessly confusing. Thanks for spotting that.

> I recommend to remove the whole paragraph on quoting and refer to RFC 4180 instead (with a note that COMMA is replaced by HTAB to specify TSV).

Disagree. That paragraph may be reworded, but replacing by a mere reference will almost guarantee that people will not bother to lookup the RFC 4180. (This is already happening for references to the LinkML and XSD specifications, which is another point I want to address in a future revision of the spec.)

> And a note should be added that the character | cannot be expressed in SSSOM/TSV for repeatable fields.

Before doing that, I think we should explore whether we can add a specific escaping mechanism.

nichtich:
[...]

PR #507, opened 2026-02-10
Allow encoding pipe characters in multi-valued slots.

gouttegd:
Resolves [#429]

This PR proposes to add to the specification of the SSSOM/TSV format an escape mechanism that allows to encode pipe characters (`|`) within the values of multi-valued slots, without interfering with the function of that character as the value separator.

This is strictly speaking a breaking change, because if a set in SSSOM/TSV format happens to contain a multi-valued slot like the following:

```
foo\|bar
```

then a SSSOM 1.0 parser will interpret that as a list containing the two values `foo\` and `bar`, whereas a SSSOM 1.1 parser that implements the mechanism proposed here will interpret that as a list containing only the single value `foo|bar`.

I am personally not much concerned by that, since I believe the likelihood that a value may happen contain a trailing backslash should be quite low.

Still, open to the idea of adding a further statement that the escaping mechanism MUST NOT be enabled when reading a SSSOM 1.0 set (i.e. a set without a `sssom_version` slot).


src/docs/spec-formats-tsv.md
```diff
@@ -83,6 +81,15 @@ Within the mappings block, the following quoting rules, adapted from [RFC 4180](
 
 SSSOM/TSV parsers MUST strip any enclosing double quotes and escaping double quotes before passing the parsed objects to the application code.
 
+#### Multi-valued slots
+
+Multi-valued slots MUST be serialised as a list of values separated by `|` characters.
+
+If a value within a multi-valued slot contains a `|` (“pipe”) character, that MUST be escaped by prepending a `\` (“backslash”) character in front of it. Likewise, if a value contains a `\` character, it MUST be escaped by prepending another `\` character in front of it if (1) the `\` character is immediately followed by a `|` character, or (2) the `\` character is the last character of the value and the value is not the last value within the slot.
```
matentzn: I am a bit concerned about the conditional interpretation of `\`. I would prefer to simply say:

- `\` MUST be escaped as `\\`
- `|` MUST be escaped as `\|`

Also I dont know if you want to add a note that the parser _has_ to read / resolve left to right to avoid a case like this:

`"abc\\|def"` (real values in memory). Now lets write this right to left:

1. `|` → `\|`
2. `\` before `|` must also be escaped

We get this:

```
abc\\\|def
```

Now lets read this left to right:

1. `\\` → `\`
2. `\|` → `|`

We end up with `"abc\|def"` (so we lost a backslash).

So in summary I would say:

1. Interpretation always left to right
2. Two rules:
   1. `\` MUST be escaped as `\\` (regardless where it is)
   2. `|` MUST be escaped as `\|`

matentzn (changes requested):

src/docs/spec-formats-tsv.md
gouttegd: I agree that the conditional escaping of `\` (the fact that it must be escaped only when needed) makes things more difficult, but it is there to reduce the impact of the change.

Without conditional escaping, an isolated `\` within a value (which for now is completely valid) would become invalid (any `\` must be escaped), for no real reason.

I don’t mind simplifying the rules for **TSV writers** (i.e. a TSV writer MUST always unconditionally escape a `\` character), but **TSV readers** SHOULD still accept an isolated (non-escaped) `\` character as a normal `\` character and not treat that as an error.

As for the direction of parsing, it didn’t occur to me to explicitly specify that because who in their right mind would try to parse a string from the end rather than from the beginning?? But I don’t mind making that explicit if it makes you feel better.

src/docs/spec-formats-tsv.md
```diff
@@ -83,6 +81,15 @@ Within the mappings block, the following quoting rules, adapted from [RFC 4180](
 
 SSSOM/TSV parsers MUST strip any enclosing double quotes and escaping double quotes before passing the parsed objects to the application code.
 
+#### Multi-valued slots
+
+Multi-valued slots MUST be serialised as a list of values separated by `|` characters.
+
+If a value within a multi-valued slot contains a `|` (“pipe”) character, that MUST be escaped by prepending a `\` (“backslash”) character in front of it. Likewise, if a value contains a `\` character, it MUST be escaped by prepending another `\` character in front of it.
```
matentzn, line 88: - I am wondering if we need some whitespace guidance, like is `a | b` the same as `a|b`? so should `a` be materialised with a trailing space in that case?

src/docs/spec-formats-tsv.md
```diff
@@ -83,6 +81,15 @@ Within the mappings block, the following quoting rules, adapted from [RFC 4180](
...
+
+Conversely, SSSOM/TSV parsers MUST interpret a `\|` sequence as a `|` character that is part of the current value, not as a value separator. Likewise, a `\\` sequence MUST be interpreted as a single `\` character. Any other occurrence of a `\` character (i.e. an occurrence that is not followed by a `|` character or another `\` character) SHOULD be interpreted as a normal `\` character. Both `\|` and `\\` sequences MUST be interpreted in the order in which they appear in the value (for example, a `\\|` MUST be read as an escaped `\` character followed by a non-escaped `|` character).
```
matentzn: I am still not convinced the ambiguity left by SHOULD is really worth it. two parsers - different interpretations. I would lean to just make this MUST.

gouttegd: I agree.

What can a parser do upon encountering an isolated `\` (by “isolated” I mean: not followed by either another `\` or a `|`)?

(a) Treat that as an error – which strictly speaking it is: since the spec says that writers MUST always escape a `\` character, finding an isolated `\` means that the data was written by a non-conforming writer.

(b) Do weird and unspecified things, such as trying to interpret more escape sequences than the ones allowed by the spec (for example, if the `\` is followed by a `n`, interpret that as a new line character).

(c) Just treat the `\` as a normal character and go on with it.

(b) would lead to even more confusion and I don’t see the point of erroring out (i.e., rejecting a mapping set) for such a simple thing, so we might as well make (c) the mandatory behaviour. Will amend the PR to that effect.

src/docs/spec-formats-tsv.md
gouttegd: > like is a | b the same as a|b?

I don’t see why that should be the same? I don’t know about SSSOM-Py’s behaviour, but SSSOM-Java preserves the whitespaces here (so the two values would be `a ` and ` b`).

gouttegd: Ping?

cthoyt (changes requested):
I'm lukewarm about doing this, but seeing some concrete examples / real use cases could easily change my mind. Can you please include that (a concrete example) in the motivation?

It's would be a pain to maintain a patchwork of string processing, so maybe we should increase the scope of this proposal to be more comprehensive on which things are allowed, and what needs to be escaped

matentzn (dismissed):
I dragged my feet a little approving this, because I prefer an outright ban on `|` character is SSSOM values (for simplicity), but @gouttegd convinced me that this is not very forward looking and user friendly.

src/docs/spec-formats-tsv.md
matentzn: Maybe you are right. I have now seen some use cases where this makes sense, but I want for the record to maintain that I was a bit on the fence here.

cthoyt:
> @cthoyt would this (invalid) example work for you instead?
> 
> ```
> #curie_map:
> #  CHEBI: http://purl.obolibrary.org/obo/CHEBI_
> #  PUBCHEM.COMPOUND: "https://identifiers.org/pubchem.compound:"
> #  chemprop: https://www.ebi.ac.uk/chebi/chebiOntology#
> #  orcid: https://orcid.org/
> #mapping_set_id: https://w3id.org/sssom/commons/examples/cxsmiles_pipe.sssom.tsv
> #mapping_set_description: "Mappings between ChEBI and PubChem Compound established via structural matching. The ferrocene entry uses a CXSMILES match_string whose pipe-delimited extension block will be incorrectly split by a parser treating | as a multivalued separator."
> #license: "https://creativecommons.org/publicdomain/zero/1.0/"
> #creator_id: 
> #  - orcid:0000-0002-7356-1779
> #mapping_provider: "https://w3id.org/sssom/core_team"
> #comment: This is an example file for SSSOM for illustration only.
> subject_id	predicate_id	object_id	mapping_justification	match_string	subject_match_field	object_match_field
> CHEBI:15365	skos:exactMatch	PUBCHEM.COMPOUND:2244	semapv:LexicalMatching	CC(=O)Oc1ccccc1C(O)=O	chemprop:CXSMILES	chemprop:CXSMILES
> CHEBI:30672	skos:exactMatch	PUBCHEM.COMPOUND:7611	semapv:LexicalMatching	c12c3c4c5c1[Fe]23451234c5c1c2c3c45 |C:4.5,0.6,1.7,2.8,3.9,7.12,6.10,9.16,10.18,8.14|	chemprop:CXSMILES	chemprop:CXSMILES
> ```
> 
> See https://www.wikidata.org/wiki/Q111662913
> 
> This is only mildly fabricated in the the sense that extended SMILES are not widely used.

Yes, thank you. A few questions/comments:

1. Is there a reason there are two mapping records in this example? I don't see any special characters in the second one
2. Official PubChem Compound CURIE prefix is `pubchem.compound` and URI prefix is `http://rdf.ncbi.nlm.nih.gov/pubchem/compound/CID`
3. Is `chemprop` real? Reference to CXSMILES would ideally be in [chemrof](https://chemkg.github.io/chemrof/) but I don't see it as a property there (suggested in https://github.com/chemkg/chemrof/issues/64). http://www.wikidata.org/entity/P10718 is the Wikidata property for CXSMILES which can be used in the meantime as `wikidata:P10718`

matentzn (approved):

matentzn:
@cthoyt example fixed according to your concerns and added to this PR

cthoyt (approved):
Optionally you might want to reference the example sssom file from the note about escaping. Thanks for being patient and following my suggestion! This really will be important for the maintainability and sustainability of the standard

PR #519, opened 2026-04-01
Add `reviewer_agreement` slot

cthoyt:
Closes #510 

- [x] `docs/` have been added/updated if necessary - see updated [confidence model](https://github.com/mapping-commons/sssom/blob/add-review-confidence/src/docs/confidence-model.md#aggregating-confidence-for-a-semantic-mapping-record) page


src/sssom_schema/schema/sssom_schema.yaml
```diff
@@ -873,6 +873,38 @@ slots:
     examples:
       - value: 0.95
         description: A confidence score of 0.95, indicating 95% confidence.
+  reviewer_confidence:
+    description: A value assigned by the reviewer of the mapping to denote their
+      confidence that the mapping record is correct. A value of 1.0 means the
+      reviewer fully agrees with the mapping record. A value of 0.0 means the
+      reviewer fully disagrees with the mapping record. A value of 0.5 means the
+      reviewer is not sure whether the mapping record is correct or not.
```
gouttegd, line 915: I note that there has been some “semantic slippage” here: what was initially intended to represent “confidence” (ranging from 0.0 aka “I have no clue whether this is correct” to 1.0 aka “I am certain this is correct”) is now used to represent “agreement” (ranging from 0.0 aka “fully disagree” to 1.0 aka “fully agree”, with the intermediate 0.5 aka “I have no clue whether to agree or not”).

Why not, but it is worth noting that this is _not_ the same kind of “confidence” as the pre-existing `confidence` slot (intended to represent the confidence of the record’s creators), where 0.0 means the creators are “fully unsure whether the record is correct or not” (equivalent to a `reviewer_confidence` of 0.5) and 1.0 means means the creators have “full confidence in the correctness of the record”.

At this point I’m half-inclined to suggest renaming this slot to something like `reviewer_agreement` to (1) reflect what the slot is really intended for and (2) avoid confusion with the the other `*confidence` slots.

gouttegd (changes requested):
No objection overall, but I’d like to at least suggest to rename the slot to something else than `_confidence` (see inline comment).

Not fully convinced that all the details about the confidence model really belong to SSSOM at all, but as long as they are not in a normative section I have no concerns with that.

src/sssom_schema/schema/sssom_schema.yaml
```diff
@@ -873,6 +873,38 @@ slots:
...
+
+      When not explicitly specified, confidence estimation algorithms should
+      consider the mapping confidence to be 1.0 by default.
```
gouttegd: I suppose this should read “consider the _reviewer_ confidence”, not the _mapping_ confidence.

src/sssom_schema/schema/sssom_schema.yaml
cthoyt: this is a good point. I am considering your suggestion and might also propose something else that is more precise/accurate than "confidence".

cthoyt: @gouttegd do you think if we rename this slot as such, then it would be alright to change the ranges to be from $[-1,1]$ instead of $[0,1]$ - I think that would also make it more clear how to use this slot

gouttegd: Good point. If the slot is no longer named `*_confidence`, then the risk of confusion with the other confidence slots becomes much lower (that’s the point), so I think it should be fine for the slot to have a wider range as well.

cthoyt: Renamed this field and updated related documentation/examples in 49336bc

src/sssom_schema/schema/sssom_schema.yaml
cthoyt: also fixed in 49336bc

gouttegd (dismissed):
I’ll let the others weigh in on the `reviewer_confidence` vs `reviewer_agremeent` naming issue and the range, but this is good for me.

src/sssom_schema/schema/sssom_schema.yaml
```diff
@@ -1304,6 +1337,21 @@ classes:
             - slot_conditions:
                 reviewer_label:
                   required: true
+      - description:
+          If a reviewer confidence is provided, then at at least one of
```
gouttegd: Minor:

```suggestion
          If a reviewer agreement value is provided, then at at least one of
```

(I don’t think there is any real ambiguity here, but just for consistency.)

gouttegd (dismissed):

src/docs/confidence-model.md
```diff
@@ -94,3 +94,116 @@ suggested to sample a minimum 50-100 semantic mappings.
...
+# curie_map:
+#   CHEBI: http://purl.obolibrary.org/obo/CHEBI_
+#   mesh: http://id.nlm.nih.gov/mesh/
+#   orcid: https://orcid.org/
+#   semapv: https://w3id.org/semapv/vocab/
+#   skos: http://www.w3.org/2004/02/skos/core#
+#   wikidata: http://www.wikidata.org/entity/
+# mapping_set_id: https://github.com/mapping-commons/sssom/blob/master/examples/schema/reviewer_agreement.sssom.tsv
+subject_id	subject_label	predicate_id	object_id	object_label	mapping_justification	mapping_tool_id	reviewer_id	reviewer_agreement
+CHEBI:10057	9H-xanthene	skos:exactMatch	mesh:C002563	xanthan gum	semapv:ManualMappingCuration	wikidata:Q111239110	orcid:0000-0003-4423-4370	-0.00
```
gouttegd:
```suggestion
CHEBI:10057	9H-xanthene	skos:exactMatch	mesh:C002563	xanthan gum	semapv:ManualMappingCuration	wikidata:Q111239110	orcid:0000-0003-4423-4370	-0.99
```

This is supposed to be an example of a reviewer that “confidently disagrees […] and denotes this by adding a low agreement (near -1.0)”.

src/sssom_schema/schema/sssom_schema.yaml
```diff
@@ -873,6 +873,38 @@ slots:
...
+    annotations:
+      added_in: "1.1"
+    see_also:
+      - https://mapping-commons.github.io/sssom/confidence-model
+      - https://github.com/mapping-commons/sssom/issues/510
+      - https://github.com/mapping-commons/sssom/pull/519
+    examples:
+      - value: 1.0
+        description:
+          A reviewer confidence of 1.0 denotes that the reviewer considers the
```
gouttegd:
```suggestion
          A reviewer agreement of 1.0 denotes that the reviewer considers the
```

Another case of a forgotten renaming following the change from `reviewer_confidence` to `reviewer_agreement`.

src/sssom_schema/schema/sssom_schema.yaml
```diff
@@ -873,6 +873,38 @@ slots:
...
+          mapping record to be correct with full confidence
+      - value: -1.0
+        description:
+          A reviewer confidence of -1.0 denotes that the reviewer considers the
```
gouttegd:
```suggestion
          A reviewer agreement of -1.0 denotes that the reviewer considers the
```

src/sssom_schema/schema/sssom_schema.yaml
```diff
@@ -873,6 +873,38 @@ slots:
...
+          mapping record to be incorrect with full confidence
+      - value: 0.0
+        description:
+          A reviewer confidence of 0.0 denotes that the reviewer is not sure
```
gouttegd:
```suggestion
          A reviewer agreement of 0.0 denotes that the reviewer is not sure
```

src/sssom_schema/schema/sssom_schema.yaml
```diff
@@ -873,6 +873,38 @@ slots:
...
+      reviewer fully agrees with the mapping record. A value of -1.0 means the
+      reviewer fully disagrees with the mapping record. A value of 0.0 means the
+      reviewer is not sure whether the mapping record is correct or not.
+
+      When not explicitly specified, confidence estimation algorithms should
+      consider the reviewer agreement to be 1.0 by default.
+    range: double
+    minimum_value: -1.0
+    maximum_value: 1.0
+    instantiates: sssom:Versionable
```
gouttegd:
```suggestion
    instantiates:
      - sssom:Versionable
```

As required by LinkML syntax (`instantiates` is multivalued, even if there is only one metaclass to instantiate it must be given as a list of one item).

Once #521 will have been merged the test suite will barf on such LinkML correctness issues.

matentzn (commented):
This looks great to me - I would approve, but would like to just make sure that @graybeal comment from slack is responded to:

> It appears that creator confidence is unchanged by reviewer agreement of 0.0, somewhat decreased by reviewer agreement of 0.25, but more decreased at 0.5? Maybe the second is an artifact, but I don't see how the first is reasonable. To know for sure I'd need to see a crisp definition of the reviewer agreement and how the scale reflects that definition.

src/sssom_schema/schema/sssom_schema.yaml
```diff
@@ -907,6 +907,38 @@ slots:
...
+    annotations:
+      added_in: "1.1"
+    see_also:
+      - https://mapping-commons.github.io/sssom/confidence-model
+      - https://github.com/mapping-commons/sssom/issues/510
+      - https://github.com/mapping-commons/sssom/pull/519
+    examples:
+      - value: 1.0
```
gouttegd:
```suggestion
      - value: "1.0"
```

To keep the LinkML validator happy.

src/sssom_schema/schema/sssom_schema.yaml
```diff
@@ -907,6 +907,38 @@ slots:
...
+        description:
+          A reviewer confidence of 1.0 denotes that the reviewer considers the
+          mapping record to be correct with full confidence
+      - value: -1.0
```
gouttegd:
```suggestion
      - value: "-1.0"
```

src/sssom_schema/schema/sssom_schema.yaml
```diff
@@ -907,6 +907,38 @@ slots:
...
+        description:
+          A reviewer confidence of -1.0 denotes that the reviewer considers the
+          mapping record to be incorrect with full confidence
+      - value: 0.0
```
gouttegd:
```suggestion
      - value: "0.0"
```

cthoyt:
> This looks great to me - I would approve, but would like to just make sure that @graybeal comment from slack is responded to:
> 
> > It appears that creator confidence is unchanged by reviewer agreement of 0.0, somewhat decreased by reviewer agreement of 0.25, but more decreased at 0.5? Maybe the second is an artifact, but I don't see how the first is reasonable. To know for sure I'd need to see a crisp definition of the reviewer agreement and how the scale reflects that definition.

I replied on Slack, I'm not sure I understood the concern but I will give time to allow for a concrete suggestion for improvement, which I would be happy to incorporate

gouttegd:
If I understand correctly, the concern is about the behaviour of the aggregation function for low reviewer agreement values (between 0.0 and 0.5). For example, with a creator’s confidence of 0.8:

| Creator confidence | Reviewer agreement | Aggregated confidence |
| ------------------ | ------------------ | --------------------- |
| 0.8                | 0.0                | 0.8                   |
| 0.8                | 0.1                | 0.775                 |
| 0.8                | 0.25               | 0.756                 |
| 0.8                | 0.5                | 0.775                 |
| 0.8                | 0.75               | 0.856                 |

(Assuming I implemented the aggregation function correctly ^^' )

It does seem a bit weird that a reviewer agreement of 0.25 yields a _lower_ aggregated confidence than a reviewer agreement of 0.1 – the reviewer is a little bit _more_ confident that the mapping is correct, yet the aggregated confidence is lower.

Another example with a creator’s confidence of 0.9:

| Creator confidence | Reviewer agreement | Aggregated confidence |
| ------------------ | ------------------ | --------------------- |
| 0.9                | 0.0                | 0.9                   |
| 0.9                | 0.1                | 0.865                 |
| 0.9                | 0.25               | 0.831                 |
| 0.9                | 0.5                | 0.825                 |
| 0.9                | 0.75               | 0.881                 |

I think the concern is valid, but I am not frankly concerned by it. OK, the proposed aggregation function behaves weirdly in some cases and therefore maybe it is not the silver bullet that will solve all confidence-related questions, but I don’t think it is being advertised as such. It is proposed as _one_ way of aggregating the creator’s confidence with the reviewer’s agreement. We do not say that it is _the_ way of doing it.

And more importantly, the proposed function is not in the _normative_ part of the SSSOM documentation. It is not part of the spec.

What _is_ part of the spec is how to interpret the values of the `reviewer_agreement` slot. And on that front I do think the spec is clear enough:

* -1.0 means the reviewer fully disagrees with the mapping record (or to say it differently, is fully confident the mapping is not correct);
* 0.0 means the reviewer has no clue;
* 1.0 means the reviewer fully agrees with the mapping record (or to say it differently, is fully confident the mapping is correct).

As far as I am concerned that’s all that matters.

gouttegd (dismissed):

gouttegd (approved):

matentzn (approved):

cthoyt:
@gouttegd thanks for the good summary. I think we're ready to merge this, and can follow-up by improving the non-normative docs with any more feedback

PR #526, opened 2026-04-04
Allow using TeX markup in documentation.

### LINKED ISSUE #525: Documentation system does not render TeX-style math
It seems that whatever we are using to generate the website does not support the use of TeX-style equations.

This is how @cthoyt ’s latest additions to the doc is rendered:

```html
<p>$$f(c,r)=(1−\left| r \right|)\times c+\left| r \right|\times \frac{r + 1}{2}$$</p>
<p>with creator confidence ($c$) and reviewer agreement ($r$). The
$\frac{r + 1}{2}$ term reweights the agreement score to work better on the
$[0,1]$ range.</p>
```

The `$` markers are simply copied over to the HTML code without any attempt at offering a way to interpret the TeX markup inside them. So, unsurprisingly, a browser simply displays the code as it is:

Image

gouttegd:
Resolves [#525]

This PR configure mkdocs to allow the use of the MathJax library for math rendering.

Not particularly fond of having to fetch a piece of Javascript from God knows where, but this is apparently the normal (at least, "normal in the world of Javascript") way of doing it [1].

closes #525

[1] https://squidfunk.github.io/mkdocs-material/reference/math/#configuration

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -23,10 +23,15 @@ markdown_extensions:
   - pymdownx.keys
   - pymdownx.mark
   - pymdownx.tilde
+  - pymdownx.arithmatex:
+      generic: true
   - admonition
   #- pymdownx.emoji:
   #    emoji_index: !!python/name:material.extensions.emoji.twemoji
   #    emoji_generator: !!python/name:material.extensions.emoji.to_svg
+extra_javascript:
+  - javascripts/mathjax.js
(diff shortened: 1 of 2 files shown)


matentzn (approved):
Bit of a pity that JS file needs to be added (would have expected the mkdocs plugin to fully handle such rendering) but if that's what is needed I have no objections

cthoyt (approved):
Image

confirming this works

PR #527, opened 2026-04-04
Re-enable OWL generation

cthoyt:
This pull request re-enables generating owl in the `make gen-project` command.

It's not documented, but I assume the reason that OWL was excluded was there were upstream bugs in the LinkML implementation that were causing it to fail. I reported this as a bug in https://github.com/linkml/linkml/issues/3358 and sent a pull request that fixes it in https://github.com/linkml/linkml/pull/3359.

This PR includes the output of the OWL generation after applying the patch to LinkML

Note: the upstream PR was merged promptly, but tests in this PR won't pass until the upstream linkml repository gets released to PyPI containing that fix

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -48,12 +48,10 @@ deploy: all mkd-gh-deploy
 # generates all project files
 gen-project: $(PYMODEL)
 	$(RUN) gen-project \
-		--exclude owl \
 		-d $(DEST) $(SOURCE_SCHEMA_PATH) && mv $(DEST)/*.py $(PYMODEL)
 
 test: validate-schema
 	$(RUN) gen-project \
-		--exclude owl \
 		-d tmp $(SOURCE_SCHEMA_PATH) 
 	$(RUN) pytest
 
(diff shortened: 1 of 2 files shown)


matentzn:
@cthoyt I need to do this - this is a little more complex. I will try to handle it next week.

cthoyt:
@matentzn this is super low priority, it's just something I noticed was weird that wasn't changing during the re-generation

matentzn:
Sorry I thought this was something else. Why do we need a OWL rendering of the SSSOM schema?

cthoyt:
🤷 it seems to be something built into LinkML output. One could also ask "why do we need protobuf"

either way, there _is_ something that's already been output before, and it's awfully out of date. so we should either update it, or delete it.

matentzn:
> either way, there is something that's already been output before, and it's awfully out of date. so we should either update it, or delete it.

I would prefer if we get rid of the OWL representation of the schema for now.

cthoyt:
@matentzn okay, deleted this file in https://github.com/mapping-commons/sssom/pull/539 as an alternative

PR #524, opened 2026-04-04
Switch from poetry to uv

cthoyt:
Closes #397

In preparation for SSSOM 1.1 release, this PR finally replaces poetry with uv. Now, I am finally able to use all of the commands in the makefile!

- [x] `docs/` have been added/updated if necessary
- [x] `make test` has been run locally
- [x] tests have been added/updated (if applicable)
- [ ] ~[CHANGELOG.md](https://github.com/mapping-commons/sssom/blob/master/CHANGELOG.md) has been updated.~
- [x] `make all`, `make docs`, `make build` also runs.

This PR also does the following:

1. Removes an `__init__.py` that was in the wrong place
2. Removes a broken test file that doesn't reflect anything about SSSOM
3. Reformats the contributing Markdown file w/ prettier and fixes minor issues (along with updating mentions of poetry)

PR #528, opened 2026-04-04
Fix docs CI action

cthoyt:
follow-up to #524: adds missing installation of uv and explicit python version

### DIFF
--- a/.github/workflows/pypi-publish.yaml
+++ b/.github/workflows/pypi-publish.yaml
@@ -16,11 +16,8 @@ jobs:
           # Checkout the code including tags required for dynamic versioning
           fetch-depth: 0
 
-      - name: Set up Python ${{ matrix.python-version }}
-        run: uv python install ${{ matrix.python-version }}
-
-      - name: Install dependencies
-        run: uv sync --all-groups
+      - name: Set up Python
+        run: uv python install 3.10
 
       - name: Build source and wheel archives
         run: uv build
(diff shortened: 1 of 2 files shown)


gouttegd (changes requested):
While we’re at it the same thing should be done to the [pypi-publish](https://github.com/mapping-commons/sssom/blob/master/.github/workflows/pypi-publish.yaml) workflow as well?

.github/workflows/pypi-publish.yaml
```diff
@@ -16,11 +16,11 @@ jobs:
...
 
-      - name: Set up Python ${{ matrix.python-version }}
-        run: uv python install ${{ matrix.python-version }}
+      - name: Install uv
+        uses: astral-sh/setup-uv@cec208311dfd045dd5311c1add060b2062131d57 # v8.0.0
 
-      - name: Install dependencies
-        run: uv sync --all-groups
+      - name: Set up Python
+        run: uv python install 3.10
```
cthoyt, line 23: i'm actually not sure installing python is necessary for this action

matentzn (commented):
I have zero objections to such changes provided they work - I cant see anything problematic, but I would have thought the uv python install is not necessary - `uv sync` will automatically download and install an appropriate Python interpreter based on our pyproject.toml's requires-python. See first paragraoh in https://docs.astral.sh/uv/guides/install-python/

gouttegd (approved):

cthoyt:
@matentzn you have to give a positive review so we can merge

matentzn (approved):

PR #531, opened 2026-04-08
Add sssom-pydantic to front page

cthoyt:
This PR adds a link to https://github.com/cthoyt/sssom-pydantic to the front page. It also slightly extends the sssom-py description to make (some of) the differences more visible


gouttegd (approved):

matentzn (approved):
I am generally very open for listing everything and I am happy to list sssom-pydantic on this page but I will warn you that in a future revision of the page, I might request some changes to the exact phrasing here as I would like to publish "canonical" pydantic classes as part of the SSSOM release soon; it might be confusing to see them both but I guess we will get there when we get there.

PR #532, opened 2026-04-08
Regenerate project files

cthoyt:
This PR regenerates the project files using the `make gen-project` target. I mistakenly thought this was part of the release workflow. In a follow-up PR, I will implement something to make sure that any change of the underlying schema YAML is also accompanied by a regeneration of the project files

- [ ] ~`docs/` have been added/updated if necessary~
- [ ] ~tests have been added/updated (if applicable)~
- [ ] ~[CHANGELOG.md](https://github.com/mapping-commons/sssom/blob/master/CHANGELOG.md) has been updated.~


gouttegd:
> I will implement something to make sure that any change of the underlying schema YAML is also accompanied by a regeneration of the project files

Please don’t. This is a very bad idea as it will cause needless conflicts whenever two PRs change the schema: even if the changes to the schema are not themselves conflicting, the schema-derived generated files are almost bound to conflict, forcing the author of one PR to go through a perfectly avoidable conflict resolution once the other PR is merged first.

Better to keep the generated files as they are during a development cycle and re-generate them once and for all shortly before a release (including a beta release if needed).

gouttegd (approved):

matentzn (approved):

cthoyt:
@matentzn can we get a new alpha release that includes the regenerated files?

PR #533, opened 2026-04-10
Publish all spec versions

### LINKED ISSUE #530: Normative part of the website should be versioned
As we head towards SSSOM 1.1, the fact that the website is not _versioned_ is going to pose a problem. In fact, I’d argue it already does.

We update the website on https://mapping-commons.github.io/sssom/ whenever we push anything to the _master_ branch (through the `deploy_documentation.yml` workflow). As a result, the website always reflects the state of the latest development version of SSSOM (the tip of the _master_ branch).

This may be fine for the _documentation_ part of the website (the non-normative part). But for the normative part (the SSSOM specification itself), I believe this is problematic.

Right now, even though officially there is no other version of the SSSOM specification than the 1.0 version published in August 2024, the only available specification document describes slots and enumeration values that do _not_, in fact, belong to SSSOM 1.0 but that will instead be part of the upcoming SSSOM 1.1. Not only is that needlessly confusing, it also means that people have no easy way of getting the actual SSSOM 1.0 specification. They can only do that by going to the _repository_ (not the website) and navigating to the SSSOM 1.0 tag – hardly a satisfying solution.

When SSSOM 1.1 will be released, then for a while the website will contain the accurate specification for the latest released version of the standard, but as soon as we will start updatig the master branch again (e.g. to add new slots for for a future 1.2 version), the same problem will happen again: the only available specification document will describe a work-in-progress and not the latest published version of the standard.

I believe the website, or at least the part of the website containing the actual specification, should be _versioned_.

That is, right now (before the SSSOM 1.1 release), it should contain two versions of the spec:

* the 1.0 version (frozen as it was in August 2024),
* and the development version (updated whenever the _master_ branch is updated).

After the SSSOM 1.1 release, it should contain:

* the 1.0 version (still frozen in the state it was in August 2024),
* the 1.1 version (frozen in the state it was on the day the 1.1 version is released),
* and the development version (which for a while will actually be identical to 1.1, until development resumes towards 1.2).

And so on as we publish newer versions.

gouttegd:
Resolves [#530]

This PR implements what has been proposed in #530: A website that shows, not just the current development version of the specification, but all previously released versions (currently only 1.0, since that’s the only release we have so far) as well.

This is done by:

(A) Moving all the normative pages of the website into `src/docs/spec` (instead of leaving them in `src/docs`). Strictly speaking this is not entirely necessary, but this allows to make a clear distinction between what part of the website is the actual spec, and what part is non-normative documentation. This is turn makes it easier to show different versions of the normative part on the same website.

(B) For every past release X.Y (so, for now, only `1.0`), we check out a Git submodule under `releases/vX.Y`, pointing to a tag or branch corresponding to the X.Y release. Then we add a new Make target `gendoc-vX.Y` that (1) copies all files from `releases/vX.Y/src/docs/spec/` (that is, all the normative files from the X.Y version) into `$(DOCDIR)/spec/v1.0` and (2) generates the LinkML-derived documentation from `releases/vX.Y/src/sssom_schema/schema/sssom_schema.yaml` (the version of the schema as it was in the X.Y release) and puts them in `$(DOCDIR)/src/v1.0`.

The `gendoc` Make target will produce a website containing only the development version of the spec. To build a website containing _all_ versions, use the new `gendoc-full` target.

The end result is that, at any time:
* `https://mapping-commons.github.io/sssom/spec/` will point to the current development version of the spec;
* `https://mapping-commons.github.io/sssom/spec/vX.Y` will point to version X.Y.

Because [cool URIs don’t change](https://www.w3.org/Provider/Style/URI), and since until now all the pages from the spec were directly under `https://mapping-commons.github.io/sssom/`, the PR also adds all the necessary redirects to ensure that any URI pointing to an old location remains valid.


gouttegd:
Reviewers: if you want to test what the “versioned“ website would look like with this PR:

```console
git fetch origin
git checkout origin/publish-all-spec-versions
git submodule init
git submodule update
make gendoc-full serve
```

Then navigate to `https://localhost:8000/sssom`.

src/docs/getting-started.md
```diff
@@ -93,7 +93,7 @@ we first expand it to `http://purl.obolibrary.org/obo/MP_0000111`.
 The `predicate_id` specifies the mapping relation between subject and object.
 Any predicate identifier may be used, but if you are just getting started, it is
 best to stick to the
-[common predicates](spec-model.md#allowed-and-common-mapping-predicates). The
+[common predicates](spec/dev/model.md#allowed-and-common-mapping-predicates). The
```
matentzn, line 96: `/dev`?

mkdocs.yml
```diff
@@ -37,20 +50,32 @@ nav:
...
+              - OWL/RDF serialisation: spec/formats-owl.md
+          - Support functions:
+              - Introduction: spec/support.md
+              - Hashing mapping records: spec/support-hashing.md
+              - Applying Chaining Rules: spec/chaining-rules.md
+      - Version 1.0:
+          - Introduction: spec/v1.0/intro.md
+          - Data model:
+              - Introduction: spec/v1.0/model.md
+              - Applying Chaining Rules: spec/v1.0/chaining-rules.md
```
matentzn, line 72: Not entirely sure this every should have been grouped under "Data model" but.. ok

pyproject.toml
```diff
@@ -40,6 +40,7 @@ tests = [
 docs = [
     "mkdocs-material==9.0.0",
     "mkdocs-mermaid2-plugin==1.1.1",
+    "mkdocs-redirects"
```
matentzn, line 43: Maybe pin this as well just to be in the habit?

matentzn (commented):
For me locally images didnt render: 

Image

My main problem is that its now very easy to go to the "wrong docs" (it does not matter what we mean by wrong here).

1. the search indexes all sites equally without discrimination - and therefore becomes useless.
2. There is no visual distinction on the pages of one version to which version it pertains (like on, say, w3org specs etc)

Why not use https://squidfunk.github.io/mkdocs-material/setup/setting-up-versioning/?

mkdocs.yml
gouttegd: It’s under “Data model” because [you put it there](https://github.com/mapping-commons/sssom/pull/377)… and [I noticed](https://github.com/mapping-commons/sssom/pull/377#discussion_r1698382781) at the time that it did not seem to have anything to do with the model.

In the current (dev) version it’s under the new “Support functions” section, but that section did not exist for version 1.0.

pyproject.toml
gouttegd: I’d need a better reason that just “to be in the habit”.

I am against pinning in general unless it is justified.

matentzn: My opinion on this is super weak and could be overruled by a mild breeze: pinning development dependencies (like docs) seems to help keeping things running smoothly; i am totally against any kind of client facing runtime dependency pinning of course. Anywyas, up to you.

src/docs/getting-started.md
gouttegd: Yes, `/dev/`. This page is in the non-versioned part of the website. From there, unless we have a specific reason (for example we are specifically talking about a precise version) we should always point to the latest (dev) version.

mkdocs.yml
matentzn: :D look at me! So dumb.

src/docs/getting-started.md
matentzn: ok.

gouttegd:
> the search indexes all sites equally without discrimination - and therefore becomes useless.

That’s a valid point, but I don’t see how versioning the entire website (as would be the case with https://squidfunk.github.io/mkdocs-material/setup/setting-up-versioning/, compared to versioning only the normative section) would change anything to that?

> Why not use https://squidfunk.github.io/mkdocs-material/setup/setting-up-versioning/?

This would create completely different websites for _everything_ (not only the normative section, but all the non-normative documentation and other pages – e.g. credits, funding, etc. – as well). That seems completely overkill to me.

matentzn:
> This would create completely different websites for everything (not only the normative section, but all the non-normative documentation and other pages – e.g. credits, funding, etc. – as well). That seems completely overkill to me.

I would not say its overkill. 80% of the users will use the dev site (which is default), and they can search it normally. the 20% will switch to the 1.0 version only if they care to look up stuff on the spec - they wont care about the outdated stuff.

An the "overkill" here has multiple dimensions, why not use something out of the box rather than build it custom? Not convinced.

gouttegd:
> > Why not use https://squidfunk.github.io/mkdocs-material/setup/setting-up-versioning/?
> 
> […] That seems completely overkill to me.

Of course I will not admit that this has nothing to with the fact that I didn’t know about mkdocs-material versioning feature. :p

gouttegd:
> why not use something out of the box rather than build it custom? Not convinced.

Fine. :) Might as well close this PR then, because any solution based on https://squidfunk.github.io/mkdocs-material/setup/setting-up-versioning/ would be so different that there is no point in using that PR as a basis.

PR #534, opened 2026-04-11
Specify standard SSSOM hashing algorithm

### LINKED ISSUE #436: Define a standard hashing procedure

gouttegd:
(This is an alternative to #529. The core proposition is the same, but the hashing procedure uses the FNV64 condensation function and hexadecimal encoding, rather than SHA2-256 and ZBase32 – as tentatively proposed in [this comment](https://github.com/mapping-commons/sssom/pull/529#issuecomment-4226360072).)

Resolves [#436]

This PR adds a new section to the normative part of the website, intended for specifying stuff that are not part of the SSSOM data model nor of any of the SSSOM serialisation formats, but that should still be formally specified so that they can be implemented in the expected way by the various implementations.

The existing page about "chaining rules" is moved to that new section, as it does not really belong to the "data model".

Finally, the core of the PR is that a new page is added to the new section to define a standard SSSOM hashing function.

closes #436

### DIFF
--- a/src/docs/spec-intro.md
+++ b/src/docs/spec-intro.md
@@ -8,7 +8,7 @@ It is divided in three sections covering the three different components of the s
 * the specification for the [serialisation formats](spec-formats.md), to read, write, and exchange SSSOM mapping sets;
 * the specification for [supporting functions](spec-support.md) to help manipulating SSSOM mappings and mapping sets.
 
-Both sections are _normative_.
+All three sections are _normative_.
 
 ## Conventions used in this document
 
--- a/src/docs/spec-intro.md
+++ b/src/docs/spec-intro.md
@@ -2,10 +2,11 @@
 
 This document is the official specification for the SSSOM standard.
 
-It is divided in two sections covering the two different components of the standard:
+It is divided in three sections covering the three different components of the standard:
 
 * the specification for the [data model](spec-model.md), to manipulate SSSOM mappings and mapping sets in your programs;
-* the specification for the [serialisation formats](spec-formats.md), to read, write, and exchange SSSOM mapping sets.
+* the specification for the [serialisation formats](spec-formats.md), to read, write, and exchange SSSOM mapping sets;
+* the specification for [supporting functions](spec-support.md) to help manipulating SSSOM mappings and mapping sets.
 
 Both sections are _normative_.
 
--- a/src/docs/spec-support-hashing.md
+++ b/src/docs/spec-support-hashing.md
@@ -6,23 +6,22 @@ and defined below.
 
 ## Rationale and purpose
 
-The SSSOM hashing function defined here is intended to allow mapping records to
-be quickly compared and, in particular, to quickly determine whether two records
-are identical.
+The SSSOM hashing function defined here allows to compute a value derived from a
+mapping record in such a way that, if two mapping records yield the same value,
+the records are highly likely to be identical.
 
 The function is intended for **interoperability** between SSSOM implementations.
 Its point is to ensure that one can always compute the same hash for the same
 mapping record regardless of which SSSOM implementation is used.
 
-When an implementation needs to assert whether two records are identical **for
-its own internal purposes** (for example, to store records into a hash table),
-it may use whatever method is best suited without regard for the SSSOM hashing
-function.
+When an implementation needs to compute a record hash **for its own internal
+purpose** (for example, to store records into a hash table), it may use whatever
+method is best suited without regard for the SSSOM hashing function.
 
 ## Hashing procedure
 
 The general principle of the SSSOM hashing function is to compute a
-ZBase32-encoded SHA2-256 hash of a canonical S-expression representing the
+hexadecimal-encoded FNV64 hash of a canonical S-expression representing the
 mapping record.
 
 ### Step 0: Propagate all condensed slots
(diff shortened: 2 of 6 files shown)


cthoyt (dismissed):

src/docs/spec-intro.md
```diff
@@ -2,10 +2,11 @@
...
 
-It is divided in two sections covering the two different components of the standard:
+It is divided in three sections covering the three different components of the standard:
 
 * the specification for the [data model](spec-model.md), to manipulate SSSOM mappings and mapping sets in your programs;
-* the specification for the [serialisation formats](spec-formats.md), to read, write, and exchange SSSOM mapping sets.
+* the specification for the [serialisation formats](spec-formats.md), to read, write, and exchange SSSOM mapping sets;
+* the specification for [supporting functions](spec-support.md) to help manipulating SSSOM mappings and mapping sets.
 
 Both sections are _normative_.
```
matentzn: All three or the first two?

src/docs/spec-support-hashing.md
```diff
@@ -0,0 +1,267 @@
...
+
+1.  Start with `(7:mapping(`.
+2.  Iterate over all slots of the `Mapping` class, in the order in which they
+    are [listed](../Mapping/#slots) in the LinkML model. Exclude the `record_id`
+    slot and the `mapping_cardinality` slot. For all other slots:
+    1. If the slot has no value for the mapping record to hash, skip to next
+       slot.
+    2. Append to the S-expression `(N:SLOTNAME`, where _SLOTNAME_ is the LinkML
+       name for the slot and _N_ is the length of the slot name (so, for
+       rxample, `(10:subject_id`, `(9:author_id`, `(10:confidence`, etc.).
```
matentzn: example

matentzn (dismissed):
This is incredibly thorough. And the discussion was a masterpiece of Open Data coordination. THANKS!

src/docs/spec-support-hashing.md
cthoyt: there are several examples below, or do you mean for something that's not covered?

src/docs/spec-intro.md
gouttegd: Good catch, thanks.

cthoyt:
confirming the FNV64 variant works in the sssom-pydantic implementation c.f. https://github.com/cthoyt/sssom-pydantic/pull/75/changes/5a2863ca63086734b5a8ce296119c64126c1c4ef

src/docs/spec-support-hashing.md
```diff
@@ -0,0 +1,267 @@
...
+       one value in the mapping record to hash):
+        1. Append `(`.
+        2. Sort the list of values in lexicographical order and iterate over the
+           sorted values. For each value _V_, append `N:V`, where _N_ is the
+           length of _V_.
+        3. Append `)`.
+    4. If the slot is typed as a floating point number (e.g. `confidence`),
+       convert the value to a string _V_ by truncating the floating point number
+       to up to 3 digits after the decimal point, rounding the last digit half
+       away from zero. Then append `N:V`, where _N_ is the length of _V_.
```
gouttegd: I am now wondering if the phrasing “rounding half away from zero” is clear enough. I think I want to add a small explanatory paragraph here, and probably a reference to IEEE 754 §4.3.1.

src/docs/spec-support-hashing.md
```diff
@@ -0,0 +1,267 @@
...
+    4. Append `))`.
+4.  Append `))`.
+
+Converting extension values to string:
+
+| Declared extension type | Conversion to string                                                                                       |
+| ----------------------- | ---------------------------------------------------------------------------------------------------------- |
+| `xsd:string`            | No conversion needed, use the value directly                                                               |
+| `xsd:integer`           | Base 10 representation of the integer value                                                                |
+| `xsd:double`            | Truncate the number to up to 3 digits after the decimal point, round the last digit half away from zero    |
```
gouttegd: Likewise, I think an explicit reference to the IEEE 754 rounding rules would make that clearer.

cthoyt: maybe worth having a few examples with different confidences that would demonstrate this?

src/docs/spec-support-hashing.md
gouttegd: No, he meant the typo: “example” was written “rxample”. ;)

src/docs/spec-support-hashing.md
```diff
@@ -111,6 +112,34 @@ Encode the hash computed in step 2 into uppercase hexadecimal, also known as
...
+
+1. The non-fractional part MUST NOT be omitted, even it is zero. For example,
+   `0.7` MUST NOT be written as `.7`.
+2. The fractional part MUST be truncated to _up to_ 3 digits _as needed_. If the
+   fractional part can be written in less than 3 digits, then it MUST NOT be
+   right-padded with zeros. For example, `0.7` MUST NOT be written as `0.700`.
+3. If the fractional part needs to be truncated, the value MUST be rounded to
+   the nearest value representable with 3 digits, rounding half away from zero.
+   This corresponds to the _roundTiesToAway_ mode as defined by [IEEE 754-2019
+   §4.3.1](https://doi.org/ 10.1109/IEEESTD.2019.8766229).
```
cthoyt: broken space in the DOI URL

```suggestion
   §4.3.1](https://doi.org/10.1109/IEEESTD.2019.8766229).
```

src/docs/spec-support-hashing.md
```diff
@@ -0,0 +1,296 @@
...
+    4. Append `))`.
+4.  Append `))`.
+
+Converting extension values to string:
+
+| Declared extension type | Conversion to string                                                                                       |
+| ----------------------- | ---------------------------------------------------------------------------------------------------------- |
+| `xsd:string`            | No conversion needed, use the value directly                                                               |
+| `xsd:integer`           | Base 10 representation of the integer value                                                                |
+| `xsd:double`            | Apply the rules from the section [Formatting floating-point values](#formatting-floating-point-values)     |
```
cthoyt, line 94: also `xsd:float`

src/docs/spec-support-hashing.md
gouttegd: Doh! 🤦

src/docs/spec-support-hashing.md
gouttegd: `xsd:float` is not among the [list of supported extension types](https://mapping-commons.github.io/sssom/spec-model/#restrictions-on-the-values-of-extension-slots) – only `xsd:double`.

cthoyt (approved):

matentzn (approved):

cthoyt:
@gouttegd :shipit:

PR #536, opened 2026-04-16
Mention the Mapping Sameness Identifier.

### LINKED ISSUE #535: Mention mapping sameness identifier
The [manual section on record idefntieirs](https://mapping-commons.github.io/sssom/record-identifiers/) now links to [the SSSOM hashing function](https://mapping-commons.github.io/sssom/spec-support-hashing/) under "Content-derived identifiers". It should also reference [Mapping Sameness Identifier](https://ts4nfdi.github.io/mapping-sameness-identifier/) below. Both are content-derived but the first identifiers the mapping record and the second only the mapping without metadata. Mapping Sameness Identifier will be used as main method to check equivalence of mappings in NFDI mapping services.

gouttegd:
Resolves [#535]

This PR adds a reference to the [Mapping Sameness Identifier](https://ts4nfdi.github.io/mapping-sameness-identifier) proposed standard, and recommends that SSSOM implementations SHOULD support it.

It also adds a mention to that standard in the non-normative guidance about minting record identifiers, to explain that it is _not_ suitable as a record identifier.

closes #535.

### DIFF
--- a/src/docs/record-identifiers.md
+++ b/src/docs/record-identifiers.md
@@ -153,7 +153,7 @@ Of note, the
 a proposed SSSOM-independent standard to compute an unique identifier for a
 mapping, is _not_ suitable as a SSSOM record identifier, because it may not be
 unique for any given record (on the contrary, it is explicitly designed to be
-identical for all records that have the same subject, the same predicate, and
+identical for all records that have the same subject, the same predicate (incl. possible negation modifier), and
 the same object, regardless of all other metadata).
 
 ##### Content-derived identifiers considered harmful
--- a/src/docs/record-identifiers.md
+++ b/src/docs/record-identifiers.md
@@ -148,6 +148,14 @@ While the resulting value may appear meaningless, and not different from a
 randomly picked number, it represents a non-opaque identifier nonetheless
 because the value is still directly dependent on the content of the record.
 
+Of note, the
+[Mapping Sameness Identifier](https://ts4nfdi.github.io/mapping-sameness-identifier/),
+a proposed SSSOM-independent standard to compute an unique identifier for a
+mapping, is _not_ suitable as a SSSOM record identifier, because it may not be
+unique for any given record (on the contrary, it is explicitly designed to be
+identical for all records that have the same subject, the same predicate, and
+the same object, regardless of all other metadata).
+
 ##### Content-derived identifiers considered harmful
 
 As noted at the very beginning of this page, the SSSOM specification is
--- a/src/docs/spec-support.md
+++ b/src/docs/spec-support.md
@@ -1,7 +1,12 @@
 # SSSOM supporting functions
 
-This section defines functions and behaviours that SSSOM implementations
-should support to help users manipulate SSSOM mapping sets.
+This section defines functions and behaviours that SSSOM implementations should
+support to help users manipulate SSSOM mapping sets.
 
-* [Hashing mapping records](spec-support-hashing.md)
-* [Chaining rules](chaining-rules.md)
+- [Hashing mapping records](spec-support-hashing.md)
+- [Chaining rules](chaining-rules.md)
+
+In addition, SSSOM implementations SHOULD also support the proposed
+[Mapping Sameness Identifier](https://ts4nfdi.github.io/mapping-sameness-identifier/)
+standard; that is, implementations SHOULD offer a way to obtain the Mapping
+Sameness Identifier for a given SSSOM mapping record.


src/docs/record-identifiers.md
```diff
@@ -148,6 +148,14 @@ While the resulting value may appear meaningless, and not different from a
 randomly picked number, it represents a non-opaque identifier nonetheless
 because the value is still directly dependent on the content of the record.
 
+Of note, the
+[Mapping Sameness Identifier](https://ts4nfdi.github.io/mapping-sameness-identifier/),
+a proposed SSSOM-independent standard to compute an unique identifier for a
+mapping, is _not_ suitable as a SSSOM record identifier, because it may not be
+unique for any given record (on the contrary, it is explicitly designed to be
+identical for all records that have the same subject, the same predicate, and
+the same object, regardless of all other metadata).
```
cthoyt, line 157: Also includes the predicate modifier

gouttegd: Yes, I saw that, but I consider that a detail that, _in the context of this warning_, can easily be glossed over. :) The point is, the Mapping Sameness Identifier does not cover the entire record and therefore is not guaranteed to be unique across all records in a set.

src/docs/spec-support.md
```diff
@@ -1,7 +1,12 @@
...
-should support to help users manipulate SSSOM mapping sets.
+This section defines functions and behaviours that SSSOM implementations should
+support to help users manipulate SSSOM mapping sets.
 
-* [Hashing mapping records](spec-support-hashing.md)
-* [Chaining rules](chaining-rules.md)
+- [Hashing mapping records](spec-support-hashing.md)
+- [Chaining rules](chaining-rules.md)
+
+In addition, SSSOM implementations SHOULD also support the proposed
```
matentzn, line 9: Two options: Either we have a lengthier debate over this, or we change SHOULD to OPTIONAL. I think SHOULD  is too strong here: either we support sameness identifiers properly in the spec or we make it optional. (I have no problem to weave this in a little stronger).

src/docs/record-identifiers.md
```diff
@@ -148,6 +148,14 @@ While the resulting value may appear meaningless, and not different from a
 randomly picked number, it represents a non-opaque identifier nonetheless
 because the value is still directly dependent on the content of the record.
 
+Of note, the
+[Mapping Sameness Identifier](https://ts4nfdi.github.io/mapping-sameness-identifier/),
+a proposed SSSOM-independent standard to compute an unique identifier for a
+mapping, is _not_ suitable as a SSSOM record identifier, because it may not be
+unique for any given record (on the contrary, it is explicitly designed to be
+identical for all records that have the same subject, the same predicate, and
```
matentzn:
```suggestion
identical for all records that have the same subject, the same predicate (incl. possible negation modifier), and
```

src/docs/record-identifiers.md
matentzn: I also stumbled across that so would favour to just add the word.

matentzn (commented):
My preferences but nothing to die for.

src/docs/spec-support.md
gouttegd: > either we support sameness identifiers properly in the spec

What do you mean by “supporting sameness identifiers _properly_ in the spec”?

As I see it, there is nothing for SSSOM to “support” here, properly or not. The Sameness Mapping Identifier is merely a function that implementations can choose to offer, the same way that they can offer the "SSSOM standard hashing function”.

nichtich: I welcome any adoption of mapping sameness identifier, but what does "SSSOM implementations SHOULD support" mean? The mapping sameness identifier is not part of the SSSOM data model and it is not going to be stored in SSSOM files because, it must be *derived* from a mapping. It can be used to "manipulate SSSOM mapping sets" but only if your use case requires finding same mappings. 

Similar argument applies to Chaining Rules: these are not strictly connected to SSSOM and only relevant if your use case involves mapping inference.

I think only Hashing mapping records is a SHOULD for SSSOM, but chaining rules and mapping sameness are optional.

matentzn: Does it not sound somehow inconsistent to (1) not really discuss the role of MSI in the spec and (2) still recommend tool developers to support it? Its like saying: tool developers should implement word count for mapping sets. (I am deliberately a bit exaggerating).

gouttegd: Again, what is there really to discuss? MSI has _no role_ in the spec, the same way that the “standard SSSOM hashing function” has no role either.

In both cases, it’s simply a function that provides a feature that SSSOM users might find useful, which is why we recommend that implementations should offer it.

matentzn: Ok :) I am trusting your judgement as I am typically sitting on the other side of this conversation.

matentzn (approved):

cthoyt (approved):

nichtich (approved):

src/docs/spec-support.md
gouttegd: @nichtich Somehow I missed your earlier comment, sorry about that.

> what does "SSSOM implementations SHOULD support" mean?

It simply means that a SSSOM library SHOULD offer, in addition to all that is necessary to manipulate SSSOM files and data (e.g. SSSOM/TSV parsing and writing functions, validating functions, etc.), a function (or a “method”, a “subroutine”, a “named procedure”, however those things are called in the language in which the implementation is written) to compute the Mapping Sameness Identifier from a SSSOM record.

> The mapping sameness identifier is not part of the SSSOM data model and it is not going to be stored in SSSOM files

Absolutely, but while the data model and the associated data formats are of course the most important part of “SSSOM“, the spec is allowed to cover slightly more than that.

Saying that “SSSOM implementations SHOULD support X” (where _X_ is not something that is part of the data model) means that we believe _X_ to be useful enough for SSSOM users that, if you claim to be a “SSSOM library”, you should provide that feature to your users, rather than asking them to implement it themselves on top of your library.

Now as for Why SHOULD rather than MAY:

I believe that, for a feature that is not part of the data model, MAY is pointless. There are, after all, MANY things that a SSSOM library could do that are not part of the data model. And if the developers of a SSSOM library want to provide additional features, they will do so anyway, because why shouldn’t they?

What would be the point of saying “SSSOM implementations MAY provide a function to derive the Mapping Sameness Identifier”? If the developers believe MSI is useful, they will provide a function to derive it whether the spec tell them they can do it or not.

PR #539, opened 2026-04-22
Remove OWL schema RDF artifact

cthoyt:
Alternative to #527


matentzn (approved):
This is stale and from my POV a useless file for SSSOM. I am inclined to delete it. @gouttegd ?

gouttegd (approved):
I don't even understand what was the point of this file.

Then again, LinkML generates a lot of files whose purpose i don't understand... :D

PR #548, opened 2026-05-23
Add new slot `derived_from`

### LINKED ISSUE #537: [New metadata element]: derived_from
Subsumes https://github.com/mapping-commons/sssom/issues/231.

## Element ID (e.g. creator_id, mapping_tool_version)

```
derived_from
```

## Value data type (e.g. URI, URL, text, xsd:boolean)

```
EntityReference
```

## Description

I would like to link derived mappings (e.g., whose justifications are `semapv:MappingChaining` or `semapv:MappingInversion`) to the mappings from which they are derived. 

I propose to do this with a multivalued slot named `derived_from` that is filled with CURIEs that reference other core mappings (explicitly, not the full SSSOM records). In many places, this will probably use CURIE-encoded [Mapping Sameness Identifiers](https://github.com/ts4nfdi/mapping-sameness-identifier) for the mappings used. 

Open questions:

- should we require all CURIEs referenced in this field to appear in record_id? Maybe with a special exception for standard MSI?

Notes/caveats:

- The subject-predicate-object triple + modifier is needed here, not the entire SSSOM record. That's why we're using MSI instead of https://github.com/mapping-commons/sssom/pull/534.
- For example purposes, this implicitly requires us to decide on a CURIE prefix and URI prefix for Mapping Sameness Identifiers. See https://github.com/ts4nfdi/mapping-sameness-identifier/issues/2. However, in practice, I think it's fine to use whatever you want here
- Multivalued slots do not have guaranteed order. This is okay, since the number of mappings that would appear in this list is usually not so big, and the mapping chain (or even network) can be trivially reconstructed. Therefore, it might make sense to suggest sorting CURIEs in this field in lexicographic order

Benefits:

- This was one of the things we wanted to include in the NFDI mapping service
- If this becomes part of SSSOM, then SSSOM can completely subsume the SeMRA data model!

## Complete example to a SSSOM file with this element

```
# curie_map:
#   cas:  	https://commonchemistry.cas.org/detail?cas_rn=
#   CHEBI: http://purl.obolibrary.org/obo/CHEBI_
#   mesh: http://id.nlm.nih.gov/mesh/
#   orcid: https://orcid.org/
#   mapping: http://w3id.org/mapping/
# license: https://creativecommons.org/publicdomain/zero/1.0/
# mapping_set_id: https://github.com/mapping-commons/sssom/blob/master/examples/schema/derived_from.sssom.tsv
# creator_id:
#  - orcid:0000-0003-4423-4370
```

| subject_id	| subject_label | predicate_id	  | object_id	|  object_label | mapping_justification   | derived_from	| comment |
| ----------- | --------------- | ----------- | -------| -------- | ----------------------- | ------------------- | ---------- |
| mesh:C000089 | ammeline | skos:exactMatch | CHEBI:28646 | ammeline | semapv:ManualMappingCuration | | |
| CHEBI:28646 | ammeline | skos:exactMatch | cas:645-92-1 | Ammeline | semapv:ManualMappingCuration | | |
| mesh:C000089 | ammeline | skos:exactMatch | cas:645-92-1 | Ammeline	 | semapv:MappingChaining | mapping:36a1f9244ea7641a90987c82f33c25c0c13712ee8f48207b2a0825f8a4e4e26a\|mapping:bb768f0b1e1643298f4df1a381001f6ed68fcc8fff49b371f0235b51dbab9e1e | this example needs to refer to both the mappings on row 18 and 19 by mapping sameness identifier |
|  CHEBI:28646 | ammeline | skos:exactMatch | mesh:C000089 | Ammeline	| semapv:MappingInversion | mapping:36a1f9244ea7641a90987c82f33c25c0c13712ee8f48207b2a0825f8a4e4e26a | this example just needs to refer to the mapping on row 18 by the mapping sameness identifier |

cthoyt:
Resolves #537

This PR adds a new field `derived_from` that enables mapping inference (e.g., via chaining, inversion) to refer back to the mappings from which it was derived.

This was motivated by needs from the NFDI use case, and also would enable nearly full subsumption of the SeMRA data model by the SSSOM data model.

It's been a few weeks since I posted the issue outlining this proposal and got some feedback on slack, so I thought it was time to make a concrete contribution.

- [x] `docs/` have been added/updated if necessary (added inference.md)

## Implementations

- https://github.com/cthoyt/sssom-pydantic/pull/108
- https://github.com/gouttegd/sssom-java/pull/19


examples/schema/derived_from.sssom.tsv
```diff
@@ -0,0 +1,19 @@
+# curie_map:
+#   cas: https://commonchemistry.cas.org/detail?cas_rn=
+#   CHEBI: http://purl.obolibrary.org/obo/CHEBI_
+#   mesh: http://id.nlm.nih.gov/mesh/
+#   orcid: https://orcid.org/
+#   semapv: https://w3id.org/semapv/vocab/
+#   skos: http://www.w3.org/2004/02/skos/core#
+#   mapping: http://www.w3.org/2004/02/skos/core#
```
gouttegd: I know you say in the documentation of `derived_from` that the IRI prefix associated to `mapping` is “left to the curators”, but using the same prefix as `skos` seems like a very bad idea. I can only assume it’s an error.

examples/schema/derived_from.sssom.tsv
```diff
@@ -0,0 +1,19 @@
...
+# license: https://creativecommons.org/publicdomain/zero/1.0/
+# mapping_set_id: https://github.com/mapping-commons/sssom/blob/master/examples/schema/derived_from.sssom.tsv
+# creator_id:
+#  - orcid:0000-0003-4423-4370
+# see_also:
+#  - https://github.com/mapping-commons/sssom/issues/537
+subject_id	subject_label	predicate_id	object_id	object_label	mapping_justification	derived_from	comment
+mesh:C000089	ammeline	skos:exactMatch	CHEBI:28646	ammeline	semapv:ManualMappingCuration
+CHEBI:28646	ammeline	skos:exactMatch	cas:645-92-1	Ammeline	semapv:ManualMappingCuration
+mesh:C000089	ammeline	skos:exactMatch	cas:645-92-1	Ammeline	 semapv:MappingChaining	mapping:36a1f9244ea7641a90987c82f33c25c0c13712ee8f48207b2a0825f8a4e4e26a|mapping:bb768f0b1e1643298f4df1a381001f6ed68fcc8fff49b371f0235b51dbab9e1e	this example needs to refer to both the mappings on row 18 and 19 by mapping sameness identifier
```
gouttegd: On this line and the next, the value of the `mapping_justification` column starts with a spurious space character (U+0020): ` semapv:MappingChaining` instead of `semapv:MappingChaining`.

gouttegd: > this example needs to refer to both the mappings on row 18 and 19 by mapping sameness identifier

The word “row” is ambiguous in a SSSOM/TSV file. Do you count all lines as rows, all lines below the YAML header block, or all lines below the TSV header?

In this context, better to use the word “line”, I think.

Regardless of the word used, line 18 is the very line containing this comment, so this makes no sense. The MSIs on that line should refer to the mappings on lines 16 and 17.

examples/schema/derived_from.sssom.tsv
```diff
@@ -0,0 +1,19 @@
...
+CHEBI:28646	ammeline	skos:exactMatch	mesh:C000089	Ammeline	 semapv:MappingInversion	mapping:36a1f9244ea7641a90987c82f33c25c0c13712ee8f48207b2a0825f8a4e4e26a	this example just needs to refer to the mapping on row 18 by the mapping sameness identifier
```
gouttegd: Same as above. This should refer to “line 16“, not “row 18”, because (1) “row” is ambiguous and (2) regardless how you count rows, 18 is incorrect.

project/graphql/sssom_schema.graphql
```diff
@@ -94,6 +94,7 @@ type Mapping
     similarityMeasure: String
```
gouttegd, line 94: As a general rule, I don’t think it is a good idea to commit all the schema-derived files as part of the PR that introduces a change to the schema.

It just makes the PR bigger for no good reason, and it creates a risk of perfectly avoidable conflicts if another PR attempts to change the schema at the same time.

Better to commit the schema-derived files once and for all shortly before a release.

src/docs/inference.md
```diff
@@ -0,0 +1,43 @@
...
+| mesh:C000089 | ammeline      | skos:exactMatch | CHEBI:28646  | ammeline     | semapv:ManualMappingCuration |                                                                                                                                                    |                                                                                                  |
+| CHEBI:28646  | ammeline      | skos:exactMatch | cas:645-92-1 | Ammeline     | semapv:ManualMappingCuration |                                                                                                                                                    |                                                                                                  |
+| mesh:C000089 | ammeline      | skos:exactMatch | cas:645-92-1 | Ammeline     | semapv:MappingChaining       | mapping:36a1f9244ea7641a90987c82f33c25c0c13712ee8f48207b2a0825f8a4e4e26a\|mapping:bb768f0b1e1643298f4df1a381001f6ed68fcc8fff49b371f0235b51dbab9e1e | this example needs to refer to both the mappings on row 18 and 19 by mapping sameness identifier |
```
gouttegd: Again, “row 18 and 19” is ambiguous/wrong here. In this context it is probably better to say something like “the first and second mapping in this table”.

src/docs/inference.md
```diff
@@ -0,0 +1,43 @@
...
+| CHEBI:28646  | ammeline      | skos:exactMatch | mesh:C000089 | Ammeline     | semapv:MappingInversion      | mapping:36a1f9244ea7641a90987c82f33c25c0c13712ee8f48207b2a0825f8a4e4e26a                                                                           | this example just needs to refer to the mapping on row 18 by the mapping sameness identifier     |
```
gouttegd: “the first mapping in this table”

src/docs/spec-model.md
```diff
@@ -311,3 +311,4 @@ Not all changes can be annotated thusly in the LinkML model, though. For changes
 * All slots that were typed as `xsd:anyURI` have been re-typed as `sssom:NonRelativeURI`. When parsing a SSSOM 1.0 set, implementations SHOULD accept relative URI values in those slots.
 * The `curation_rule` and `curation_rule_text` slots which previously only existed on the `Mapping` class, have been added to the `MappingSet` class. Both slots have now been typed [propagatable](#propagation-of-mapping-set-slots).
 * A new value `0:0` has been added to the `mapping_cardinality_enum`.
+* Added the `derived_from` field to the `Mapping` class to reference other mappings from which it was derived (see [#537](https://github.com/mapping-commons/sssom/issues/537))
```
gouttegd: This is not needed. This section is intended for changes to the model that cannot be inferred solely by the `added_in` annotations. That is not the case for the addition of a complete new slot.

src/sssom_schema/schema/sssom_schema.yaml
```diff
@@ -1144,6 +1144,43 @@ slots:
       - https://github.com/mapping-commons/sssom/issues/78
       - https://github.com/mapping-commons/sssom/pull/259
       - https://github.com/mapping-commons/sssom/blob/master/examples/schema/issue_tracker.sssom.tsv
+  derived_from:
+    description:
+      A set of [mapping sameness identifiers
+      (MSIs)](https://github.com/ts4nfdi/mapping-sameness-identifier) for
+      mappings that were used to derive this mapping, e.g., through mapping
+      chaining or inversion, represented as CURIEs. It's left to the curator to
```
gouttegd: Not with the current MSI specification. The spec is clear that the MSI is _exactly_ `mapping:<hash>[~]`. Any prefix declaration that associates `mapping:` to anything else than `mapping:` itself would produce something that may _look like_ a MSI in CURIE form, but whose value is not in fact a valid MSI.

gouttegd (commented):
Mostly looks good to me, but before this can be merged, either https://github.com/ts4nfdi/mapping-sameness-identifier/issues/2 must be resolved or the PR must be amended to remove the bit about the appropriate CURIE/URI prefix for encoding MSIs as CURIEs being “left to the curators”.

examples/schema/derived_from.sssom.tsv
```diff
@@ -0,0 +1,19 @@
+# curie_map:
```
gouttegd, line 1: Since this file uses a slot that did not exist in SSSOM 1.0, it must explicitly indicate the version of SSSOM that it is compliant with:

```
sssom_version: 1.1
```

cthoyt: fixed in 71161a3

src/docs/spec-model.md
cthoyt: fixed in 7c06437

src/docs/inference.md
cthoyt: fixed in 812e09a

src/docs/inference.md
cthoyt: fixed in 812e09a

project/graphql/sssom_schema.graphql
cthoyt: good point. I removed these and squashed the history

examples/schema/derived_from.sssom.tsv
cthoyt: updated text in fcbf056

examples/schema/derived_from.sssom.tsv
cthoyt: updated text in fcbf056

examples/schema/derived_from.sssom.tsv
cthoyt: good catch, fixed in d93e27c. we should follow-up with a separate PR that adds some linting to these example files

examples/schema/derived_from.sssom.tsv
cthoyt: I updated this in 77007f7. I am also removing most references to MSI from this proposal because I don't want to get blocked by the need to update MSI spec upsteam (Jakob has been unavailable for comment and has made some changes without discussion that as you pointed out, made it harder to reconcile with this SSSOM propsal)

src/sssom_schema/schema/sssom_schema.yaml
cthoyt: Removed reference to MSI in the description of this slot in 31f7cbb. Now it just says a CURIE should be used to represent each S-P-O-predicate modifier quad, and does not specify how this should be done.

The examples still reference MSI

cthoyt:
@gouttegd thanks for the feedback. I agree that the introduction of MSIs here is problematic because of current unresolved discussions about MSI, so opted to remove references to it in the description.

I also switched to using example.com dummy URI prefixes in the examples

src/sssom_schema/schema/sssom_schema.yaml
```diff
@@ -1144,6 +1144,43 @@ slots:
       - https://github.com/mapping-commons/sssom/issues/78
       - https://github.com/mapping-commons/sssom/pull/259
       - https://github.com/mapping-commons/sssom/blob/master/examples/schema/issue_tracker.sssom.tsv
+  derived_from:
+    description:
+      A set of CURIEs referring to mappings (i.e.,
+      subject-predicate-object-predicate-modifier quadruple) that were used to
+      derive this mapping, e.g., through mapping chaining or inversion,
+      represented as CURIEs.
```
gouttegd: ”A set of CURIEs […] represented as CURIEs” sounds… weird.

I would rather describe the slot as a set of _identifiers_ (which, as with all other `EntityReference`-typed slots, are represented as CURIEs in the SSSOM/TSV format).

cthoyt: oop fixed in 5ef24ec

gouttegd (dismissed):
Looks good to me.

cthoyt:
@matentzn ready for your final review

src/docs/inference.md
```diff
@@ -0,0 +1,47 @@
...
+  `semapv:BackgroundKnowledgeBasedMatching` as a justification
+
+## Referring to Evidence
+
+The `derived_from` field was introduced in
+[#537](https://github.com/mapping-commons/sssom/issues/537) in order to
+reference the original subject-predicate-object-predicate modifier quadruple
+from which new mappings are inferred/derived.
+
+The following example demonstrates how the `derived_from` field when doing both
```
matentzn:
```suggestion
The following example demonstrates how the `derived_from` field can be leveraged when doing both
```

missing verb

src/docs/inference.md
```diff
@@ -0,0 +1,47 @@
...
+| CHEBI:28646  | ammeline      | skos:exactMatch | cas:645-92-1 | Ammeline     | semapv:ManualMappingCuration |                                                                                                                                                    |                                                                                                       |
+| mesh:C000089 | ammeline      | skos:exactMatch | cas:645-92-1 | Ammeline     | semapv:MappingChaining       | mapping:36a1f9244ea7641a90987c82f33c25c0c13712ee8f48207b2a0825f8a4e4e26a\|mapping:bb768f0b1e1643298f4df1a381001f6ed68fcc8fff49b371f0235b51dbab9e1e | this example needs to refer to the first two mappings in this table by mapping sameness identifier    |
+| CHEBI:28646  | ammeline      | skos:exactMatch | mesh:C000089 | Ammeline     | semapv:MappingInversion      | mapping:36a1f9244ea7641a90987c82f33c25c0c13712ee8f48207b2a0825f8a4e4e26a                                                                           | this example just needs to refer to the first mappingin this table by the mapping sameness identifier |
```
matentzn:
```suggestion
| CHEBI:28646  | ammeline      | skos:exactMatch | mesh:C000089 | Ammeline     | semapv:MappingInversion      | mapping:36a1f9244ea7641a90987c82f33c25c0c13712ee8f48207b2a0825f8a4e4e26a                                                                           | this example just needs to refer to the first mapping in this table by the mapping sameness identifier |
```

src/sssom_schema/schema/sssom_schema.yaml
```diff
@@ -1144,6 +1144,43 @@ slots:
...
+      - https://github.com/mapping-commons/sssom/issues/537
+      - https://github.com/mapping-commons/sssom/pull/548
+      - https://github.com/mapping-commons/sssom/blob/master/examples/schema/derived_from.sssom.tsv
+    examples:
+      - value: mapping:36a1f9244ea7641a90987c82f33c25c0c13712ee8f48207b2a0825f8a4e4e26a
+        description:
+          A single value can be used when inverting the mapping
+          mesh:C000089-skos:exactMatch-CHEBI:28646, represented with the
+          [mapping sameness identifier
+          (MSI)](https://github.com/ts4nfdi/mapping-sameness-identifier) for is
```
matentzn:
```suggestion
          (MSI)](https://github.com/ts4nfdi/mapping-sameness-identifier)
```

src/docs/inference.md
```diff
@@ -0,0 +1,47 @@
...
+[#537](https://github.com/mapping-commons/sssom/issues/537) in order to
+reference the original subject-predicate-object-predicate modifier quadruple
+from which new mappings are inferred/derived.
+
+The following example demonstrates how the `derived_from` field when doing both
+mapping chaining (i.e., using multiple mappings to infer a new one) and mapping
+inversion. The local unique identifiers used for mappings in this example are
+related to the proposal in
+https://github.com/ts4nfdi/mapping-sameness-identifier (which currently is under
+finalization). For now, the SSSOM specification isn't currently prescribing how
```
matentzn:
```suggestion
review). For now, the SSSOM specification isn't currently prescribing how
```

src/docs/inference.md
```diff
@@ -0,0 +1,47 @@
...
+  `semapv:MappingInversion` as a justification
+- Inference via prior knowledge, which should be tagged with
+  `semapv:BackgroundKnowledgeBasedMatching` as a justification
+
+## Referring to Evidence
+
+The `derived_from` field was introduced in
+[#537](https://github.com/mapping-commons/sssom/issues/537) in order to
+reference the original subject-predicate-object-predicate modifier quadruple
+from which new mappings are inferred/derived.
```
matentzn, line 41: would be good to  have a reference for what a "mapping quad" is and why its not simple a mapping triple. But I leave it to you.

src/docs/inference.md
matentzn: I would like to see a single sentence verbalisation of the example. Someone entirely new to this might still not get what "derived from" really means.

src/sssom_schema/schema/sssom_schema.yaml
```diff
@@ -1144,6 +1144,43 @@ slots:
       - https://github.com/mapping-commons/sssom/issues/78
       - https://github.com/mapping-commons/sssom/pull/259
       - https://github.com/mapping-commons/sssom/blob/master/examples/schema/issue_tracker.sssom.tsv
+  derived_from:
+    description:
+      A set of identifiers referring to mappings (i.e.,
+      subject-predicate-object-predicate-modifier quadruple) that were used to
+      derive this mapping, e.g., through mapping chaining or inversion,
+      represented as CURIEs.
```
matentzn: What irks me a little that this wording precludes me from using "record_id" as a value for this field. I am not sure what you have already discussed regarding that, but I dont see exactly why it should not be allowed at all?

matentzn (commented):
Agree in principle with all this, but some quick fixes and RFCs

src/sssom_schema/schema/sssom_schema.yaml
gouttegd: The way I read it, now that any reference to MSI has been removed, you _can_ actually put records ID there..

A record ID identifies _more_ than a “mapping quadruple”, but a mapping quadruple is part of what is identified by a record ID.

To me this is in line with many other aspects of SSSOM, in that we do not prescribe much. You can use a MSI, a record ID, or any other way you can think to point to a quadruple (or to something that contains at least a quadruple).

src/docs/inference.md
cthoyt: Added an explanation of difference between triple, quad, and mapping record in the inference.md in 0d86489

src/docs/inference.md
cthoyt: Added a verbalization of the examples in 0d86489

src/sssom_schema/schema/sssom_schema.yaml
cthoyt: I added a bit more context about why mapping quad is preferred (but not prescribing this) in 0d86489

cthoyt:
@matentzn @gouttegd ready for re-review

cthoyt:
sorry nico, I deleted a comment about fixing the word `CURIEs` in the text that was resolved already but seemed to be hanging for some reason (github bug?)

src/docs/inference.md
```diff
@@ -0,0 +1,125 @@
...
+2. Full mapping records are inflexible: the SSSOM data should be flexible so if
+   additional evidence (i.e., records) for a given mapping quadruple are found,
+   then the confidence in the inferred/derived mapping (e.g., chained or
+   inverted)can be adjusted accordingly. This is possible because most chaining
+   and inversion algorithms logically operate on mapping quadruples, and not on
+   records.
+
+Note: the local unique identifiers used for mappings in this example are related
+to the proposal in https://github.com/ts4nfdi/mapping-sameness-identifier (which
+currently is under review). For now, the SSSOM specification isn't currently
```
gouttegd: Nitpick: “For now […] currently” is redundant.

src/docs/inference.md
```diff
@@ -0,0 +1,125 @@
...
+prescribing how to assign CURIEs to mapping quads.
```
gouttegd: I’d rather say “assign _identifiers_ to mapping quads”.

Yes, the identifiers would be represented as CURIEs in SSSOM/TSV, but not necessarily in other formats (typically RDF).

src/sssom_schema/schema/sssom_schema.yaml
```diff
@@ -1144,6 +1144,43 @@ slots:
       - https://github.com/mapping-commons/sssom/issues/78
       - https://github.com/mapping-commons/sssom/pull/259
       - https://github.com/mapping-commons/sssom/blob/master/examples/schema/issue_tracker.sssom.tsv
+  derived_from:
+    description:
+      A set of identifiers referring to mappings (i.e.,
+      subject-predicate-object-predicate modifier quadruple) that were used to
+      derive this mapping, e.g., through mapping chaining or inversion,
+      represented as CURIEs.
```
gouttegd: The fact that the identifiers are represented as CURIEs is a SSSOM/TSV-specific detail that does not need to appear here. The specification already mandates elsewhere that all `EntityReference`-typed slots MUST be represented as CURIEs in SSSOM/TSV.

This is misleading as it suggests that this slot in particular should _always_ be represented using CURIEs regardless of the serialisation format, which is not the case.

gouttegd (changes requested):
A few last details, but otherwise still good.

src/docs/inference.md
cthoyt: ece1535

src/sssom_schema/schema/sssom_schema.yaml
cthoyt: ece1535

src/docs/inference.md
cthoyt: ece1535

gouttegd (dismissed):

src/docs/inference.md
```diff
@@ -0,0 +1,125 @@
...
+| CHEBI:10057 | 9H-xanthene   | skos:exactMatch | Not                | mesh:C002563 | xanthan gum  | semapv:ManualMappingCuration |                                                                                                                                                     |                                                                                                                            |
+| CAS:92-83-1 | Xanthane      | skos:exactMatch |                    | CHEBI:10057  | 9H-xanthene  | semapv:ManualMappingCuration |                                                                                                                                                     |                                                                                                                            |
```
matentzn: CAS upper case - should be lower

src/docs/inference.md
```diff
@@ -0,0 +1,125 @@
...
+| CAS:92-83-1 | Xanthene      | skos:exactMatch | Not                | mesh:C002563 | xanthan gum  | semapv:MappingChaining       | mapping:58f24ccfaf71431276da873c9e7b77ea61a2425e4e8b283b943542290deb292b~\|mapping:bb1162fb2afb1c519c0aa8be98c352061720af220e2d052c571a1fecabff9800 | this example uses the first two mappings in the table, importantly, incorporating the negative modifier in the identifiers |
```
matentzn: CAS upper case, should be lower

src/docs/inference.md
matentzn: Something is wrong with the label here: its Xanthane, while the next row has the same ID and Xanthene label..

src/docs/inference.md
```diff
@@ -0,0 +1,125 @@
...
+For the purposes of inference, the mapping quadruple should be used:
+
+1. Mapping triples are insufficient: without the judgment of whether a mapping
+   is true or false, then an algorithm could accidentally conclude from
+   `A skos:exactMatch B` and `B (not) skos:exactMatch C` that
+   `A skos:exactMatch C`. This is why mapping triples are insufficient
+2. Full mapping records are inflexible: the SSSOM data should be flexible so if
+   additional evidence (i.e., records) for a given mapping quadruple are found,
+   then the confidence in the inferred/derived mapping (e.g., chained or
+   inverted)can be adjusted accordingly. This is possible because most chaining
```
matentzn:
```suggestion
   inverted) can be adjusted accordingly. This is possible because most chaining
```

src/docs/inference.md
```diff
@@ -0,0 +1,125 @@
...
+   records.
+
+Note: the local unique identifiers used for mappings in this example are related
+to the proposal in https://github.com/ts4nfdi/mapping-sameness-identifier (which
+currently is under review). For now, the SSSOM specification isn't prescribing
+how to assign identifiers to mapping quadruples.
+
+### Example with Negative Mappings
+
+The following example uses mapping chaining combine with negated mappings to
```
matentzn:
```suggestion
The following example uses mapping chaining combined with negated mappings to
```

src/docs/inference.md
```diff
@@ -0,0 +1,125 @@
...
+
+The `derived_from` field was introduced in
+[#537](https://github.com/mapping-commons/sssom/issues/537) in order to
+reference the original subject-predicate-object-predicate modifier quadruple
+from which new mappings are inferred/derived.
+
+The following example demonstrates how the `derived_from` field can be leveraged
+in two scenarios:
+
+1. mapping chaining. the table contains a SKOS exact match from `mesh:C000089`
```
matentzn:
```suggestion
1. mapping chaining. The table contains a SKOS exact match from `mesh:C000089`
```

src/docs/inference.md
```diff
@@ -0,0 +1,125 @@
...
+   to `CHEBI:28646` and from `CHEBI:28646` to `cas:645-92-1`. The third row of
+   the table contains a SKOS exact match from `mesh:C000089` to `cas:645-92-1`
+   produced through mapping chaining. The `derived_from` column in this row
+   contains CURIEs referring to the
+   `mesh:C000089`-`skos:exactMatch`-`CHEBI:28646`-True and
+   `CHEBI:28646`-`skos:exactMatch`-`cas:645-92-1`-True mapping quadruples,
+   concatenated with a pipe
+2. mapping inversion. the fourth row of the table contains a SKOS exact match
```
matentzn:
```suggestion
2. mapping inversion. The fourth row of the table contains a SKOS exact match
```

src/docs/inference.md
```diff
@@ -0,0 +1,125 @@
...
+Note: the local unique identifiers used for mappings in this example are related
+to the proposal in https://github.com/ts4nfdi/mapping-sameness-identifier (which
+currently is under review). For now, the SSSOM specification isn't prescribing
+how to assign identifiers to mapping quadruples.
+
+### Example with Negative Mappings
+
+The following example uses mapping chaining combine with negated mappings to
+infer a non-trivial negative mapping. This illustrates why mapping quadruples
+(i.e., subject-predicate-modifier-object) are required over mapping triples
```
matentzn:
```suggestion
(i.e., subject-predicate-object-predicate modifier) are required over mapping triples
```

matentzn (changes requested):
Some nits but basically done

src/docs/inference.md
cthoyt: fixed in c555652

src/docs/inference.md
cthoyt: fixed in c555652, it should have been Xanthene. Getting careless with manual writing here...

cthoyt: fixed in c555652

matentzn (approved):
Great work! Thank you!

gouttegd (approved):

PR #538, opened 2026-04-22
Use `mike` to versionise the website.

gouttegd:
Resolves [#530]

- [ ] ~~`docs/` have been added/updated if necessary~~
- [ ] ~~`make test` has been run locally~~
- [ ] ~~tests have been added/updated (if applicable)~~
- [ ] ~~[CHANGELOG.md](https://github.com/mapping-commons/sssom/blob/master/CHANGELOG.md) has been updated.~~

This PR configures MkDocs to use [mike](https://github.com/jimporter/mike) in order to create a fully versionised website.

**Edit: Forget what’s below, unless you want for some reason to test the entire procedure for yourself. Go directly to [that comment](https://github.com/mapping-commons/sssom/pull/538#issuecomment-4303338087) if you just want to see what the result would look like.**

Full procedure for generating the versionised website for testing:

To test the result:

1. Make sure you have the `gh-pages` branch locally tracked in your local repo. If not:

```console
$ git branch -t gh-pages origin/gh-pages
```

2. Create a tag on the `gh-pages` branch (so that you can easily revert all the changes you’ll have to do on that branch in order to test the versionised website):

```console
$ git tag savepoint gh-pages
```

3. Checkout the branch corresponding to this PR:

```console
$ git branch -t versionise-entire-website origin/versionise-entire-website
$ git checkout versionise-entire-website
```

4. Wipe out the current content of the `gh-pages` branch

```console
$ uv run mike delete --all
```

5. Build the `dev` version of the documentation:

```console
$ make gendoc
```

6. Deploy the `dev` version of the documentation and set it as the default version:

```console
$ uv run mike deploy dev
$ uv run mike set-default dev
```

Now you need to build the website for version 1.0:

7. Checkout the tag corresponding to the 1.0 version:

```console
$ git checkout v1.0.0
```

8. Amend `pyproject.toml` to add `mike` to the `docs` dependency group and amend `mkdocs.yml` to add the `mike` configuration (as is done on this PR, but now you need to back port that to your local checkout of the v1.0.0 tag).

9. Build the documentation for version 1.0. Remember that we were not using `uv` yet back in 1.0, so you must make sure you still have Poetry around:

```console
$ make gendoc
```

10. Deploy the 1.0 documentation:

```console
$ poetry run mike deploy 1.0
```

11. Switch back to the `versionise-entire-website` branch (strictly speaking not necessary, but this allows to be able to use `uv` back and avoid having to use Poetry more than strictly necessary):

```console
$ git checkout versionise-entire-website
[...]

PR #549, opened 2026-05-26
Rebuild project after adding `derives_from` slot to Mapping model

cthoyt:
Following #538, this PR rebuilds all artifacts with `make gen-project`.


matentzn (approved):

gouttegd (approved):
I would normally do that only just before a release, but I have no objection (and maybe you want do to a release soon anyway?).

PR #555, opened 2026-05-28
Format markdown

cthoyt:
This PR runs `npx prettier --check --prose-wrap always --write **/*.md` to format all markdown in the repo. It doesn't change any content (except one minor typo and one section where I converted italics at the beginning of paragraphs into proper headers)

This PR does not add any linting checks to CI (yet, maybe in a follow-up)

There are many times I wanted to do typo checks/editing/revising but wasn't able to because the diff would have been too big because of using single lines. This enables me to more effectively edit and revise


src/docs/spec-support-hashing.md
```diff
@@ -49,11 +49,11 @@ The S-expression MUST be assembled as follows:
...
-        1. Append `(`.
-        2. Sort the list of values in lexicographical order and iterate over the
-           sorted values. For each value _V_, append `N:V`, where _N_ is the
-           length of _V_.
-        3. Append `)`.
+       1. Append `(`.
+       2. Sort the list of values in lexicographical order and iterate over the
+          sorted values. For each value _V_, append `N:V`, where _N_ is the
+          length of _V_.
+       3. Append `)`.
```
gouttegd: This change is incorrect. There is a reason for this additional space in front of those numbers: it’s to indicate that we are dealing with “sub-sub-items” (a list within a list within a list).

This is what allows the (correct) [existing rendering](https://mapping-commons.github.io/sssom/dev/spec-support-hashing/#step-1-turn-the-mapping-record-into-a-canonical-s-expression):

Image

This change completely messes up the rendering:

Image

src/docs/spec-support-hashing.md
```diff
@@ -77,11 +77,11 @@ The S-expression MUST be assembled as follows:
...
-        1. Append `(N:PROP`, where _PROP_ is the property identifying the
-           extension and _N_ is the length of the property.
-        2. Use the table below to transform the extension value into a string
-           _V_ based on the declared type of the extension.
-        3. Append `N:V)`, where _N_ is the length of the string value _V_.
+       1. Append `(N:PROP`, where _PROP_ is the property identifying the
+          extension and _N_ is the length of the property.
+       2. Use the table below to transform the extension value into a string _V_
+          based on the declared type of the extension.
+       3. Append `N:V)`, where _N_ is the length of the string value _V_.
```
gouttegd: Same as above.

Correct rendering (before the change):

Image

Wrong rendering (after this change):

Image

gouttegd (changes requested):

src/docs/spec-support-hashing.md
```diff
@@ -38,52 +38,53 @@ S-expression as per
...
-       2. Use the table below to transform the extension value into a string _V_
-          based on the declared type of the extension.
-       3. Append `N:V)`, where _N_ is the length of the string value _V_.
-    4. Append `))`.
-4.  Append `))`.
+1. Start with `(7:mapping(`.
+2. Iterate over all slots of the `Mapping` class, in the order in which they are
+   [listed](../Mapping/#slots) in the LinkML model. Exclude the `record_id` slot
+   and the `mapping_cardinality` slot. For all other slots:
+   1. If the slot has no value for the mapping record to hash, skip to next
```
gouttegd, line 45: This is still incorrect. Here is the rendering:

Image

This time it’s the sub-items that are not recognised as such.

gouttegd (changes requested):

PR #560, opened 2026-07-27
Fix typos `uriOrCurie` -> `Uriorcurie`

cthoyt:


src/docs/spec-support-hashing.md
```diff
@@ -278,7 +278,7 @@ Hash value:
 #    type_hint: xsd:integer
 #  - slot_name: ext_baz
 #    property: EXPROP:bazProperty
-#    type_hint: linkml:Uriorcurie
+#    type_hint: linkml:uriOrCurie
```
matentzn: Where did you get this information?

I cant figure it out either..

https://github.com/linkml/linkml-model/blob/986c7881bae48055d3dbd9106205bc86c5215df4/linkml_model/model/schema/types.yaml#L125

cthoyt: :((( actually I don't know

https://w3id.org/linkml/Uriorcurie works,

https://w3id.org/linkml/uriorcurie goes to 404 :/

cthoyt: https://linkml.io/linkml-model/latest/docs/Uriorcurie/ also uses `linkml:uriorcurie`...

gouttegd: It’s `linkml:Uriorcurie` (expanded to `https://w3id.org/linkml/Uriorcurie`).

The source of truth is the LinkML schema, which says that the type is called `uriorcurie`. The native type URI is derived from (1) prepending the `https://w3id.org/linkml/` prefix and (2) turning the name into _CamelCase_, using internal underscores (when present) to identify words boundaries.

The name transformation rules are not specified anywhere as far as I know (file that along with the many things that the LinkML “specification” really *should* specify but does not…), but they are implemented in LinkML-Py and can be relied upon to determine what the URI of any given type is.

Check all the other types defined in the `linkml:types` schema, you’ll see the same logic everywhere:

* `date` in the schema → `linkml:Date` (that’s why https://w3id.org/linkml/Date resolves while https://w3id.org/linkml/date does not)
* `datetime` in the schema → `linkml:Datetime` (likewise: https://w3id.org/linkml/Datetime works, https://w3id.org/linkml/datetime does not)
* `date_or_datetime` in the schema → `linkml:DateOrDatetime` (likewise: https://w3id.org/linkml/DateOrDatetime works, https://w3id.org/linkml/date_or_datetime does not)

So `uriorcurie` in the schema → the native URI is `linkml:Uriorcurie`.

`uriorcurie` should only be used within a LinkML schema (typically in a `range` slot). As for `uriOrCurie`, I have no idea where that name comes from, but it does not in fact exist.

gouttegd (changes requested):
The proposed change is incorrect. `linkml:Uriorcurie` **is** the correct URI for LinkML’s `uriorcurie` type.

src/docs/spec-support-hashing.md
gouttegd: > As for uriOrCurie, I have no idea where that name comes from, but it does not in fact exist.

Which makes me realise that I’ve been guilty myself of using `linkml:uriOrCurie` at some places within the SSSOM spec…

cthoyt: thanks @gouttegd. I will update this PR with fixes

src/docs/spec-model.md
```diff
@@ -25,7 +25,7 @@ Of note, within a set, a mapping may not necessarily be uniquely identified by t
...
+Throughout the model, identifiers to external resources are represented using the custom type [`EntityReference`](EntityReference.md) (based on the LinkML type [`linkml:Uriorcurie`](https://w3id.org/linkml/Uriorcurie)), which accepts both full-length IRIs and [CURIEs](https://www.w3.org/TR/curie/) as possible identifier formats. (Note however that serialisation formats may mandate the use of one identifier format over the other; for example, the [SSSOM/TSV](spec-formats-tsv.md) format requires the systematic use of CURIEs, whereas the [OWL/RDF](spec-formats-owl.md) format conversely requires the systematic use of IRIs).
```
gouttegd, line 28: Not strictly required here as “LinkML type `uriorcurie`” is both correct (`uriorcurie` is the correct name for that type) and unambiguous (so there’s no need to refer to the type by its URI).

gouttegd (dismissed):

src/docs/spec-model.md
cthoyt: given the discussion here, I wanted to be as explicit as possible

gouttegd (approved):

matentzn (approved):

PR #565, opened 2026-07-31
Add additional example of hashing with extension slots.

### LINKED ISSUE #563: Add additional examples of hashing w/ extension slots
currently, https://mapping-commons.github.io/sssom/dev/spec-support-hashing/ has an integer and a CURIE test for extension slots. it would be good to extend the example with one for a float, a boolean, a date a datetime, and a xsd:anyURI

this would help sanity check that my implementation of extension slots in https://github.com/cthoyt/sssom-pydantic/pull/149 is correct

gouttegd:
Resolves #563

Add one more example of hashing a mapping record containing missing types of extension values (float, boolean, date, datetime, URI).

### DIFF
--- a/src/docs/spec-support-hashing.md
+++ b/src/docs/spec-support-hashing.md
@@ -154,6 +154,8 @@ Given the following mapping set in SSSOM/TSV format:
 #  orcid: https://orcid.org/
 #  semapv: https://w3id.org/semapv/vocab/
 #  skos: http://www.w3.org/2004/02/skos/core#
+#mapping_set_id: https://example.org/sets/4FC65A74-A75A-4CCC-8679-907D235C9FAC
+#license: https://creativecommons.org/licenses/by/4.0/
 subject_id	predicate_id	object_id	mapping_justification	creator_id
 FBbt:00001234	skos:exactMatch	UBERON:0005678	semapv:ManualMappingCuration	orcid:0000-0000-5678-1234|orcid:0000-0000-1234-5678
 ```
@@ -194,6 +196,8 @@ resulting bytes in hexadecimal would yield the following final value:
 #  semapv: https://w3id.org/semapv/vocab/
 #  skos: http://www.w3.org/2004/02/skos/core#
 #  wikidata: https://www.wikidata.org/wiki/
+#mapping_set_id: https://example.org/sets/94AC84D6-A64D-4C96-BFDA-3487F5C1E242
+#license: https://creativecommons.org/licenses/by/4.0/
 #subject_source: KF_FOOD:DB
 #object_source: wikidata:Q55118395
 #object_source_version: http://purl.obolibrary.org/obo/foodon/releases/2022-02-01/foodon.owl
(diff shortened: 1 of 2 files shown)


src/docs/spec-support-hashing.md
```diff
@@ -296,3 +296,43 @@ Hash value:
...
+#    property: EXPROP:verified
+#    type_hint: xsd:boolean
+#  - slot_name: ext_verification_date
+#    type_hint: xsd:date
+#  - slot_name: ext_timestamp
+#    property: EXPROP:timestamp
+#    type_hint: xsd:datetime
+#  - slot_name: ext_see_also
+#    property: rdfs:seeAlso
+#    type_hint: xsd:anyURI
```
cthoyt, line 337: would you please add a `mapping_set_id` so this is a valid SSSOM document

src/docs/spec-support-hashing.md
```diff
@@ -296,3 +296,43 @@ Hash value:
...
+#    property: EXPROP:accuracy
+#    type_hint: xsd:float
+#  - slot_name: ext_verified
+#    property: EXPROP:verified
+#    type_hint: xsd:boolean
+#  - slot_name: ext_verification_date
+#    type_hint: xsd:date
+#  - slot_name: ext_timestamp
+#    property: EXPROP:timestamp
+#    type_hint: xsd:datetime
```
cthoyt: is the second t capitalized? `xsd:dateTime`

gouttegd: Good catch. SSSOM-Java got it wrong since the very beginning 😱

src/docs/spec-support-hashing.md
```diff
@@ -317,7 +317,7 @@ Hash value:
 #    type_hint: xsd:date
 #  - slot_name: ext_timestamp
 #    property: EXPROP:timestamp
-#    type_hint: xsd:datetime
+#    type_hint: xsd:dateTime
```
cthoyt, line 320: sorry to be picky but I expect that the s-expression and the hash should also change, right?

cthoyt: or not, this doesn't get reflected there.

gouttegd: Not in this case. The type hint does not appear in the S-expression. It has no reason to appear since it’s not an attribute of the mapping record – it is an attribute of the extension definition. (Same reason why the `slot_name` does not appear either; only the _property_ does appear in the S-expression, because that’s what is used to identify the extension.)

Changing the type hint _could_ sometimes result in a different S-expression, if the new type hint caused the extension value to be serialised differently.

For example, let’s say you have an extension defined with a type hint of `xsd:string` and a value of `0.12345`. Since the value is typed as a `string`, it is rendered in the S-expression exactly as it is. But let’s say that you change the type hint to `xsd:float`: now the value is typed as a floating point number, and it should be rendered in the S-expression as `0.123`. So in this example, even though the type hint itself never appears in the S-expression, changing the type hint does result in a different S-expression.

But in the example given in the docs, the value `2026-07-31T11:11:11+01:00` is rendered in the S-expression in exactly the same way regardless of whether it is typed as a `xsd:dateTime` (correctly) or as a `xsd:datetime` (incorrectly), so changing the type hint does not change the S-expression and therefore the hash.

cthoyt (approved):
confirming this all works as expected, also in my implementation. thanks @gouttegd!

matentzn (approved):
Cant find anything to complain about which is disappointing

gouttegd:
> Cant find anything to complain about which is disappointing

:D Sorry, next time I’ll make sure to introduce a mistake or two for you (but it will be up to you to catch them before @cthoyt does :p ).

